PR #56, opened 2024-01-11
Pass assigns to changeset function

Flo0807:
There is a case where we need to access a user in the changeset to validate permissions to change a particular field. In a Phoenix / LiveView application without Backpex, you could handle this case in the event handler before calling a changeset function. In Backpex we currently have no layer between the event and the changeset function. So we have to access the user via the assigns in the changeset function. We pass the `_target` to the changeset function for the same reason.

I suppose there is a better but more complex solution than passing all assigns to every function to handle such cases. Perhaps we should add more layers to hook into. We should address this in future updates!


Flo0807:
I added ignore-for-release label as we reworked this in another pull request

PR #83, opened 2024-02-01
Fix updating item on show view

Flo0807:
An error occurs if an item is updated while the show view is open. In this case we are trying to update the list of items, but there is no `items` key.

With this merge request, the `handle_info` callback takes into account the live action. It also fetches the item from the database to properly load all associations. Previously the `HasManyThrough` field value was not correctly updated by pub sub events.

### DIFF
--- a/lib/backpex/live_resource.ex
+++ b/lib/backpex/live_resource.ex
@@ -1364,7 +1364,14 @@ defmodule Backpex.LiveResource do
       end
 
       @impl Phoenix.LiveView
-      def handle_info({"backpex:" <> unquote(event_prefix) <> "deleted", item}, socket) do
+      def handle_info({"backpex:" <> unquote(event_prefix) <> "created", item}, socket)
+          when socket.assigns.live_action in [:index, :resource_action] do
+        {:noreply, refresh_items(socket)}
+      end
+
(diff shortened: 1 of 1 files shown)


lib/backpex/live_resource.ex
```diff
@@ -1364,28 +1364,27 @@ defmodule Backpex.LiveResource do
...
-          {:noreply, refresh_items(socket)}
-        else
-          {:noreply, socket}
-        end
-      end
-
-      @impl Phoenix.LiveView
-      def handle_info({"backpex:" <> unquote(event_prefix) <> "created", item}, socket) do
+      def handle_info({"backpex:" <> unquote(event_prefix) <> "created", item}, socket)
+          when socket.assigns.live_action in [:index, :resource_action] do
```
krns, line 1368: Do we need fallbacks when this (and the other) guard does not match? Or is this not a valid code path?

Flo0807: We already have a fallback `handle_info` function added by the  `__before_compile__ ` hook (see [live_resource.ex#L1509](https://github.com/naymspace/backpex/blob/develop/lib/backpex/live_resource.ex#L1509))

PR #94, opened 2024-02-07
Pass metadata to changesets via keyword list

Flo0807:
In the past we have had requests from applications where we needed additional metadata in the changeset function. For example, the `assigns` or the name of the form `target`. We solved this by supporting multiple changeset functions with different arity.

We were not happy with this solution as additional metadata would further increase the arity.

For this reason we now change the arity of all changeset functions called by Backpex to `3`. The first and second parameters contain the item and the params as you know them from Ecto changesets.

The third parameter is a keyword list containing any additional metadata passed to the changeset.

At the moment we pass the following metadata to changesets

- `:assigns` - the assigns
- `:target` - the name of the `form` target that triggered the changeset call

The keyword list allows us to easily add additional metadata without having to support another changeset function with a different arity.

As part of these changes, the changeset functions of the Item Actions and Resource Actions are also being adapted to conform to the LiveResource.

The following breaking changes will be introduced:

- Require the `changeset/3` function to be passed to the `Backpex.LiveResource`.
- Require `changeset/3` instead of `changeset/2` in `Backpex.ResourceAction`.
- Require `changeset/3` instead of `changeset/2` in `Backpex.ItemAction`.

If you do not need any metadata in your changeset, you can of course mark the param as unused with an underscore and simply dont use it.


demo/lib/demo_web/resource_actions/email.ex
```diff
@@ -30,14 +30,14 @@ defmodule DemoWeb.ResourceActions.Email do
...
-  def changeset(change, attrs) do
+  def changeset(change, attrs, _metadata \\ []) do
     change
     |> cast(attrs, @required_fields)
     |> validate_required(@required_fields)
   end
 
   @impl Backpex.ResourceAction
-  def handle(_socket, _params) do
+  def handle(_socket, params) do
```
pehbehbeh, line 40:
```suggestion
  def handle(_socket, _params) do
```

mix.exs
```diff
@@ -96,7 +96,8 @@ defmodule Backpex.MixProject do
     [
       "guides/introduction/installation.md",
       "guides/introduction/translations.md",
-      "guides/advanced/full_text_search.md"
+      "guides/advanced/full_text_search.md",
+      "guides/upgrading/v0.2.md"
```
pehbehbeh, line 100: There should also be added a "group for extras" below, I think.

```
"Upgrade Guides": ~r{guides/upgrading/.*}
```

pehbehbeh (approved):

mix.exs
Flo0807: Done ✅

PR #110, opened 2024-02-16
Improve ci pipeline

Flo0807:
Many improvements for the CI pipeline.

We test Backpex in a separate matrix with different combination of OTP and Elixir version.

Along with that, we build a testing image based on the Dockerfile (`MIX_ENV=test`, `target=runtime`) and mostly use it to test the demo application. We run all lint commands provided in `demo/package.json`, compile the demo application with `--warnings-as-errors`, audit deps and run the tests. If all (demo) testing steps are successful, a runtime image is being built (`MIX_ENV=prod`, `target=runtime`).

The images are hosted in the GitHub Container registry:  https://github.com/orgs/naymspace/packages?repo_name=backpex We use [docker/metadata-action](https://github.com/docker/metadata-action) to label and tag the images. See the [README](https://github.com/docker/metadata-action/blob/master/README.md) of the action for explainations on how images get tagged and labelled in detail.

To sum it up:
| Event               | Ref                           | Docker Tags                |
|---------------------|-------------------------------|----------------------------|
| `pull_request`      | `refs/pull/2/merge`           | `pr-2`                     |
| `push`              | `refs/heads/master`           | `master`                   |
| `push`              | `refs/heads/releases/v1`      | `releases-v1`              |
| `push tag`          | `refs/tags/v1.2.3`            | `v1.2.3`, `latest`         |
| `push tag`          | `refs/tags/v2.0.8-beta.67`    | `v2.0.8-beta.67`, `latest` |
| `workflow_dispatch` | `refs/heads/master`           | `master`                   |

This means after a push to `develop` we have these two images:
- `ghcr.io/naymspace/backpex/testing:develop`
- `ghcr.io/naymspace/backpex/runtime:develop`

After a tag push with the title `0.1.0`:
- `ghcr.io/naymspace/backpex/testing:0.1.0`
- `ghcr.io/naymspace/backpex/runtime:0.1.0`

We use [actions/cache](https://github.com/actions/cache) to cache the `_build`, `deps` and `priv/plts` folder when executing Backpex tests. The cache depends on the Elixir and OTP version combination and the `mix.lock` file.

In addition, we use a registry cache to speed up docker builds. See [Cache management with GitHub Actions](https://docs.docker.com/build/ci/github-actions/cache/#registry-cache) for explanations.

When running the Demo tests, we also use [actions/cache](https://github.com/actions/cache) to cache `priv/plts`. These tests are running inside our container. That's why we have to install `tar` because the version of tar shipped as part of busybox in our container doesn't support the `--posix` option which is used by [actions/cache](https://github.com/actions/cache)

Example pipeline:
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: CI
 
 on:
+  workflow_dispatch:
   push:
   release:
     types: [published]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,8 +227,8 @@ jobs:
 
   build-runtime:
     name: "Build runtime"
-    runs-on: ubuntu-latest
-    needs: [lint-demo, test-compile, test-demo, deps-audit]
+    runs-on: docker:git
+    # needs: [lint-demo, test-compile, test-demo, deps-audit]
 
     permissions:
       contents: read
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,11 +121,11 @@ jobs:
 
       - name: Determine the elixir version
         id: elixir_version
-        run: echo "::set-output name=version::$(grep -h elixir .tool-versions | awk '{ print $2 }' | awk -F - '{print $1}')"
+        run: echo "{version}={grep -h elixir .tool-versions | awk '{ print $2 }' | awk -F - '{print $1}'}" >> $GITHUB_OUTPUT
 
       - name: Determine the otp version
         id: otp_version
-        run: echo "::set-output name=version::$(grep -h erlang .tool-versions | awk '{ print $2 }')"
+        run: echo "{version}={grep -h erlang .tool-versions | awk '{ print $2 }'}" >> $GITHUB_OUTPUT
 
       - name: Use versions
         run: echo "Using Elixir version ${{ steps.elixir_version.outputs.version }} and OTP version ${{ steps.otp_version.outputs.version }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,35 +121,39 @@ jobs:
 
   test-demo:
     needs: build-demo-testing
-    name: "Test demo"
+    name: "Lint demo"
     runs-on: ubuntu-latest
-    container:
-      # we need to get the image by the output of the previous job because
-      # we can not acces envs here 
-      image: ${{ needs.build-demo-testing.outputs.image }}
   
     steps:
-      - name: Run hex info
-        run: |
-          mix hex.info
+      - name: "Run linter inside demo testing image"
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -85,7 +85,6 @@ SHELL ["/bin/bash", "-c"]
 COPY --from=builder /etc/bashrc.d /etc/bashrc.d
 COPY --from=builder /etc/bash.bashrc /etc/bash.bashrc
 COPY --from=builder /opt/scripts /opt/scripts
-COPY .bashrc /root/
```
Flo0807, line 88: I had to remove this line, because `.bashrc` can not be found.

.github/workflows/ci.yml
```diff
@@ -5,28 +5,297 @@ on:
...
-  ci:
+  test:
+    name: "Test backpex"
     runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
+    strategy:
+      matrix:
+        elixir: ['1.15.7', '1.16.1']
+        erlang: ['26.1.2', '26.2.1']
```
Flo0807, line 20: Feel free to add more version.

.github/workflows/ci.yml
```diff
@@ -5,28 +5,297 @@ on:
...
+      - name: Add tar that supports --posix option to be used by cache action
+        run: |
+          apk add --no-cache tar
+
+      - name: Restore the plts cache
+        uses: actions/cache@v4
+        id: plts-cache
+        with:
+          path: /opt/app/demo/priv/plts
+          key: ${{ runner.os }}-demo-plts-mixlockhash-${{ hashFiles(format('{0}{1}', github.workspace, '/opt/app/demo/mix.lock')) }}
```
Flo0807, line 170: This `key` does not seem to work properly, because the cache is always saved on the primary key `Linux-demo-plts-mixlockhash-` (see [this](https://github.com/naymspace/backpex/actions/runs/7931127493/job/21654741621) pipeline). The hash of the `mix.lock` does not get attached. 

When do we even want to invalidate the demo plts cache?

.github/workflows/ci.yml
pehbehbeh: The latest two OTP versions (25 and 26) should be fine, I think.

Dockerfile
pehbehbeh: I have moved all files that only affect the Docker image to the `.docker` subfolder.

`.docker/.bashrc` should work.

.github/workflows/ci.yml
pehbehbeh: We should remove Dialyzer. See also internal discussion.

pehbehbeh (approved):

Dockerfile
Flo0807: `.docker/root/.bashrc` works

PR #113, opened 2024-02-19
Refactor filter modules and make them reusable

Flo0807:
Refactoring of the filter modules to make them more reusable and prevent dialyzer warnings. Moved most of the functions outside the quote block.


lib/backpex/filters/boolean.ex
```diff
@@ -38,6 +38,8 @@ defmodule Backpex.Filters.Boolean do
   > In addition it will add a `render` and `render_form` function in order to display the corresponding filter.
   > It will also implement the `Backpex.Filter.query` function to define a boolean query.
   """
+  use Phoenix.Component, global_prefixes: ~w(x-)
```
pehbehbeh, line 41: Can't we `use BackpexWeb, :filter` here? `global_prefixes` are already defined there. Seems redundant in every filter module.

lib/backpex/filters/boolean.ex
```diff
@@ -48,87 +50,115 @@ defmodule Backpex.Filters.Boolean do
...
       @impl Backpex.Filter
       def render(var!(assigns)) do
-        var!(assigns) =
-          var!(assigns)
-          |> assign(:label, option_value_to_label(options(), var!(assigns).value))
+        var!(assigns) = assign(var!(assigns), :options, options())
 
         ~H"""
-        <%= @label %>
+        <BooleanFilter.render options={@options} value={@value} />
```
pehbehbeh, line 67: Can we do even less in the quote blocks? Ideally, we only call the function outside the block here ([see Credo](https://hexdocs.pm/credo/Credo.Check.Refactor.LongQuoteBlocks.html)). Then we would save one of the two Heex parts.

pehbehbeh (changes requested):

pehbehbeh (approved):

lib/backpex/filters/boolean.ex
Flo0807: I need to assign the options because they can only be accessed from within the quote block. However, I have simplified it as follows:
```elixir
@impl Backpex.Filter
def render(assigns) do
  assigns = assign(assigns, :options, options())
  BooleanFilter.render(assigns)
end
```

PR #117, opened 2024-02-20
Use `to_form` helper for forms and refactor inputs

Flo0807:
- use `to_form` helper to create forms
- refactor inputs to not rely on `phoenix_html` helper functions


lib/backpex/html/form.ex
```diff
@@ -4,152 +4,249 @@ defmodule Backpex.HTML.Form do
...
+
+  slot :inner_block
+
+  def field_input(%{field: %Phoenix.HTML.FormField{} = field} = assigns) do
+    translated_errors =
+      field.errors
+      |> Enum.map(fn error ->
+        {text, _error} = translate_form_error(error, assigns.field_options)
+        text
+      end)
```
pehbehbeh, line 53: We also have this code in `multi_select`. Maybe add a `translated_form_errors` function?

lib/backpex/html/form.ex
```diff
@@ -4,152 +4,249 @@ defmodule Backpex.HTML.Form do
...
+            id={@id}
+            name={@name}
+            value="true"
+            checked={@checked}
+            class={[
+              "toggle phx-no-feedback:toggle phx-no-feedback:toggle-primary",
+              @input_class,
+              @errors == [] && "toggle-primary",
+              @errors != [] && "toggle-error"
+            ]}
```
pehbehbeh, line 149: Maybe add a helper for this common case?

- first param: default class
- second param: no error classes
- third param: error classes

```suggestion
            class={input_classes(["toggle phx-no-feedback:toggle phx-no-feedback:toggle-primary", @input_class], "toggle-primary", "toggle-error")}
```

lib/backpex/html/resource.ex
```diff
@@ -961,13 +960,22 @@ defmodule Backpex.HTML.Resource do
   defp metric_toggle(assigns) do
     visible = Backpex.Metric.metrics_visible?(assigns.metric_visibility, assigns.live_resource)
 
-    assigns = assign(assigns, :visible, visible)
+    form =
+      to_form(
+        %{"_resource" => assigns.live_resource, "_cookie_redirect_url" => assigns.current_url},
+        as: :toggle_metrics
+      )
```
pehbehbeh, line 967:
```suggestion
    form =
      %{"_resource" => assigns.live_resource, "_cookie_redirect_url" => assigns.current_url}
      |> to_form(as: :toggle_metrics)
```

pehbehbeh (approved):
Awesome refactoring :) I only noted a few small things. You can merge on your own.

lib/backpex/html/form.ex
Flo0807: In this case, would we also need to pass the `@errors` attribute to the `input_classes` function, or how does it decide whether to include error classes or no error classes?

pehbehbeh: Right... maybe just leave it as is.

PR #193, opened 2024-04-05
Enable dynamic select options and improve docs

krns:


lib/backpex/fields/boolean.ex
```diff
@@ -4,8 +4,8 @@ defmodule Backpex.Fields.Boolean do
 
   ## Options
 
-  * `:debounce` - Optional integer timeout value (in milliseconds), or "blur".
-  * `:throttle` - Optional integer timeout value (in milliseconds).
+  * `:debounce` - Optional integer timeout value (in milliseconds), "blur" or function that receives the assignes.
+  * `:throttle` - Optional integer timeout value (in milliseconds) or function that receives the assignes.
```
Flo0807, line 8:
```suggestion
  * `:throttle` - Optional integer timeout value (in milliseconds) or function that receives the assigns.
```
see other fields as well

lib/backpex/fields/select.ex
```diff
@@ -25,7 +25,7 @@ defmodule Backpex.Fields.Select do
   @impl Backpex.Field
   def render_value(assigns) do
     options = Map.get(assigns.field_options, :options)
```
Flo0807, line 27:
```suggestion
    options = get_options(assigns.field_options)
```

In my opinion, it would be better to get the options here. We could change `assigns_options` to `get_options`. Then we do not need to change the logic of `get_label` and can leave the rest as it is.

PR #220, opened 2024-04-16
Improve pipeline

Flo0807:


.github/workflows/ci.yml
```diff
@@ -100,12 +91,6 @@ jobs:
         run: |
           mix gettext.extract --check-up-to-date
 
-      - name: Publish package on hex.pm
-        if: github.event_name == 'release'
-        env:
-          HEX_API_KEY: ${{ secrets.HEX_API_KEY }}
-        run: mix hex.publish --yes
-
```
Flo0807, line 108: We have an independent pipeline job / step for releasing the package.

.github/workflows/ci.yml
```diff
@@ -177,7 +163,7 @@ jobs:
           build-args: |
             MIX_ENV=test
     outputs:
-      image: ${{ steps.meta.outputs.tags }}
+      image: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME_TESTING }}@${{ steps.docker.outputs.digest }}
```
Flo0807, line 166: Using the sha is safer because if the image has multiple tags (e.g., `latest` and version number) we cannot use them as a list to pull the image. Then we would need to extract only one tag.

PR #222, opened 2024-04-18
Fix custom alias causes nil bindings in query

Flo0807:
If a field is found by `get_custom_alias` and no `custom_alias` is defined for that field, the function will return `nil`, which in some cases will break the query (e.g., if no `select` option is defined).


makl9797 (approved):
I added the `field_name` to the `get_custom_alias` function to get the correct field. Otherwise the filters would not work. Please check again if everything is working.

PR #227, opened 2024-04-22
Add index edit for belongs to field

Flo0807:


lib/backpex/fields/belongs_to.ex
```diff
@@ -119,6 +172,18 @@ defmodule Backpex.Fields.BelongsTo do
...
+    Map.get(field_options, :display_field_form, display_field)
+  end
+
+  defp get_options(repo, queryable, field_options, display_field, assigns) do
+    queryable
+    |> from()
+    |> maybe_options_query(field_options, assigns)
+    |> repo.all()
+    |> Enum.map(&{Map.get(&1, display_field), Map.get(&1, :id)})
+  end
```
Flo0807, line 185: Eventually, we should move this code to `resource.ex` and prevent having queries in our fields

pehbehbeh (approved):

PR #245, opened 2024-05-07
Fix resizing of the item action column for dynamically added item actions

phyr97:
If an item action is added dynamically after clicking another item action, for example, the column of the item actions is not resized, which results in an overflow of the icons. Now, the table listens for changes in its DOM tree and adjusts the width of the item action column based on the width of the element containing all the item action buttons.

### DIFF
--- a/lib/backpex/html/resource/resource_index_table.html.heex
+++ b/lib/backpex/html/resource/resource_index_table.html.heex
@@ -2,21 +2,32 @@
   class="!static table"
   x-cloak
   x-data="{
-    actionsWidth: null,
+    actionsWidth: 0,
     headerHeight: null,
     setActionsWidth() {
-      $nextTick(
-        () => {
-          document.querySelectorAll('.item-action-column')
-            .forEach((x) => {
-              this.actionsWidth = this.actionsWidth > x.clientWidth ? this.actionsWidth : x.clientWidth
-            })
-        }
-      )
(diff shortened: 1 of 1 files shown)


pehbehbeh:
@makl9797 What is the status of this PR? Please mark as draft if applicable. Please request a review when ready.

lib/backpex/html/resource/resource_index_table.html.heex
```diff
@@ -102,8 +113,8 @@
         ]}
       >
         <div
-          class="flex items-center justify-end space-x-2"
-          x-bind:style="`height: ${rowHeight - 1}px !important; width: ${actionsWidth - 32}px !important`"
```
makl9797, line 106: > @makl9797 What is the status of this PR? Please mark as draft if applicable. Please request a review when ready.

I have removed the highlighted part so that the width of the column is `auto`. Previously the width was fixed, which meant that `setActionsWidth` did not work correctly either, as the width that is set in `setActionsWidth` is based on the previously set width. The problem now is that the column is no longer displayed correctly when the page is loaded due to the missing initial width setting. The correct width is calculated and set for a short time, but then immediately overwritten again. I have not yet been able to find out why this is the case, but the problem could be solved with a delay of half a second. However, this is not a reasonable solution to the problem. I have not worked on it any further so far.

PR #297, opened 2024-06-10
handle 404 for resource items

pehbehbeh:
Resource show/edit fails when providing an incorrect `:id` in the path, because `Repo.get` just returns `nil` (for invalid UUIDs) or an cast error (for other invalid id strings).

### DIFF
--- a/lib/backpex/resource.ex
+++ b/lib/backpex/resource.ex
@@ -256,7 +256,7 @@ defmodule Backpex.Resource do
   @doc """
   Gets a database record with the given fields by the given id, possibly being enhanced by the given item_query.
   """
-  def get(assigns, item_query, fields, id) do
+  def get!(assigns, item_query, fields, id) do
     %{
       repo: repo,
       schema: schema
--- a/lib/backpex/resource.ex
+++ b/lib/backpex/resource.ex
@@ -270,8 +270,15 @@ defmodule Backpex.Resource do
     |> maybe_join(associations)
     |> maybe_preload(associations, fields)
     |> maybe_merge_dynamic_fields(fields)
-    |> where([{^schema_name, schema_name}], schema_name.id == ^id)
+    |> where_id(schema_name, id)
     |> repo.one!()
+  end
+
+  defp where_id(query, schema_name, id) do
+    case Ecto.UUID.cast(id) do
+      {:ok, valid_id} -> where(query, [{^schema_name, schema_name}], schema_name.id == ^valid_id)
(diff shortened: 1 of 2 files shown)


lib/backpex/resource.ex
```diff
@@ -271,7 +271,7 @@ defmodule Backpex.Resource do
     |> maybe_preload(associations, fields)
     |> maybe_merge_dynamic_fields(fields)
     |> where([{^schema_name, schema_name}], schema_name.id == ^id)
-    |> repo.one()
+    |> repo.one!()
```
pehbehbeh, line 274: @Flo0807 This will raise a `Ecto.NoResultsError` and show a proper 404 page instead of returning `nil` and failing later on. Should we rename `get` to `get!` or `one!`?

Flo0807: > Should we rename get to get! or one!?

yes, I like`get!`

lib/backpex/resource.ex
```diff
@@ -270,8 +270,15 @@ defmodule Backpex.Resource do
...
+    |> where_id(schema_name, id)
+    |> repo.one!()
+  end
+
+  defp where_id(query, schema_name, id) do
+    case Ecto.UUID.cast(id) do
+      {:ok, valid_id} -> where(query, [{^schema_name, schema_name}], schema_name.id == ^valid_id)
+      :error -> where(query, [{^schema_name, schema_name}], true == false)
+    end
   end
```
pehbehbeh, line 282: @Flo0807 This fixes another issue with invalid IDs when visiting `/admin/posts/INVALID/show` (for example) you would otherwise get an error like this:

```
/opt/app/lib/backpex/resource.ex:273: value `"INVALID"` cannot be dumped to type :binary_id in query:

from p0 in Demo.Post,
  as: :post,
  left_join: u1 in Demo.User,
  as: :user,
  on: p0.user_id == u1.id,
  left_join: c2 in Demo.Category,
  as: :custom_category,
  on: p0.category_id == c2.id,
  where: p0.id == ^"INVALID",
  distinct: [asc: p0.id],
  select: p0,
  preload: [{:user, #Ecto.Query<from u0 in Demo.User, as: :user, select: merge(u0, %{full_name: fragment("concat(?, ' ', ?)", u0.first_name, u0.last_name)})>}, :category, :tags]
```

The error case query `true == false` will result in a `Ecto.NoResultsError` and properly render a 404 page.

Flo0807 (approved):

PR #303, opened 2024-06-13
Elixir 1.17 and OTP 27.0

krns:


mix.exs
```diff
@@ -8,7 +8,7 @@ defmodule Backpex.MixProject do
     [
       app: :backpex,
       version: @version,
-      elixir: "~> 1.12",
+      elixir: "~> 1.15",
```
krns, line 11: @pehbehbeh @Flo0807 I think we should bump this version, but I'm not sure which version we want to "officially" support.

krns: Bumped it to 1.16

Flo0807: 👍

Flo0807:
@krns can you update this PR with the new erlang / elixir version and fix the conflict?

PR #366, opened 2024-06-18
Support other id types besides binary_id

Flo0807:
- Rewrite all migrations (we need to start with a fresh database in the demo) to change `id` type in `users` table to `bigserial` more easily
- Support other id types besides `:binary_id` (tested with `:bigserial`: default Ecto migration primary key type)
- Format schemas


Flo0807:
This must be tested carefully. There may be some side effects.

pehbehbeh:
Awesome! :-) 

Maybe we should change on of the demo resources to integer IDs?

Flo0807:
> Awesome! :-)
> 
> Maybe we should change one of the demo resources to integer IDs?

Changed `film_reviews` primary key type to bigserial

pehbehbeh:
I would have chosen a resource where we have all of the CRUD actions. Once we have tests for the demo in place, all actions would be tested with integer IDs.

Flo0807:
> I would have chosen a resource where we have all of the CRUD actions. Once we have tests for the demo in place, all actions would be tested with integer IDs.

Done. I changed `users` table primary key to `bigserial`. I rewrote all migrations to make this change. Therefore, this also contains the changes from #376 

We have to start with a fresh database in our demo application.

This is not a breaking change for Backpex, but one for the demo. Should we add https://github.com/naymspace/backpex/labels/breaking-change to this PR?

This is ready for review now 🚀

@pehbehbeh

pehbehbeh:
> We have to start with a fresh database in our demo application.

I will do that manually. Maybe we can reset the demo database on each deploy. 🤔 

> Should we add [breaking-change](https://github.com/naymspace/backpex/labels/breaking-change) to this PR?

No, not relevant to Backpex users.

PR #379, opened 2024-06-20
Bump phoenix_html to 4.1.1

guess:
The `inputs_for` and `hidden_inputs_for` functions have been moved to `phoenix_html_helpers` 

https://hexdocs.pm/phoenix_html/changelog.html#v4-0-0-2023-12-19


Flo0807:
The app doesn't start:

> app-1       | Because "your app" depends on "phoenix_html empty" which doesn't match any versions, version solving failed.
app-1       | ** (Mix) Hex dependency resolution failed
app-1 exited with code 1

This is because we also have `phoenix_html` as a dependency in our demo project, not only in Backpex itself (see `demo/mix.exs`).

I deleted the dependency from the demo project. This will start the app, but raise an error if I try to access `http://localhost:4000/admin/users/new`.

> app-1       | ** (Protocol.UndefinedError) protocol Enumerable not implemented for {:safe, []} of type Tuple. This protocol is implemented for the following type(s): CircularBuffer, DBConnection.PrepareStream, DBConnection.Stream, Date.Range, Ecto.Adapters.SQL.Stream, File.Stream, Floki.HTMLTree, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, Jason.OrderedObject, List, Map, MapSet, Phoenix.LiveView.LiveStream, Postgrex.Stream, Range, Stream
app-1       |     (elixir 1.15.7) lib/enum.ex:1: Enumerable.impl_for!/1
app-1       |     (elixir 1.15.7) lib/enum.ex:166: Enumerable.reduce/3
app-1       |     (elixir 1.15.7) lib/enum.ex:4387: Enum.filter/2
app-1       |     (backpex 0.3.2) lib/backpex/fields/has_many_through.ex:180: Backpex.Fields.HasManyThrough.render_form/1

The reason might be that we do not use the `inputs_for` function in [`Backpex.Fields.HasManyThrough#L507`](https://github.com/naymspace/backpex/blob/develop/lib/backpex/fields/has_many_through.ex#L507) as intended. The `HasManyThrough` field is still in beta state and there needs to be some refactoring done, before we can safely update to `phoenix_html` v4, I think.

guess:
Ah sorry I didn't see the demo project. I can try to take a look at the previous state and what happens on the update to see if there's any obvious fixes :)

Flo0807:
> Ah sorry I didn't see the demo project. I can try to take a look at the previous state and what happens on the update to see if there's any obvious fixes :)

That's awesome, thanks! 🙏

pehbehbeh:
I have come a bit further. We need to change [this line](https://github.com/naymspace/backpex/blob/0.3.1/lib/backpex/fields/has_many_through.ex#L507): 

```diff
-    inputs_for(form, association.pivot.field)
+    form.impl.to_form(form.source, form, association.pivot.field, [])
```

Because before we were using [this variant of `inputs_for/3`](https://github.com/phoenixframework/phoenix_html/blob/v3.3.4/lib/phoenix_html/form.ex#L632-L645) of which is now deprecated. But the change above replicates the old behaviour.

The field is still broken though...

guess:
The addresses field doesn't seem to work on the latest develop either – selecting any values has no effect.
You can also see that happening here actually: https://backpex.live/admin/users/new

Flo0807:
> The addresses field doesn't seem to work on the latest develop either – selecting any values has no effect. You can also see that happening here actually: https://backpex.live/admin/users/new

Thanks for your report. I provided a fix for this in #387

pehbehbeh:
> The addresses field doesn't seem to work on the latest develop either

Yep, thats what I meant with:

> The field is still broken though...

Thanks @Flo0807 for the fix! :-) 

Then all we need here is the update in the demo and the line change I suggested.

@guess Are you going to finish this or should I take over?

guess:
Yay! Pushed and tested that it works in the demo project 🎉 @Flo0807 @pehbehbeh

PR #388, opened 2024-06-21
Add `create_button_label` option to live resource

### LINKED ISSUE #382: Add configuration for create button label
The label for the create button should be configurable. This also includes the create button in the empty state.

- [ ] Add a `create_button_label` function / option to live resources
- [ ] Add guide for new feature

Flo0807:
closes #382


lib/backpex/live_resource.ex
```diff
@@ -121,6 +121,11 @@ defmodule Backpex.LiveResource do
   @callback return_to(socket :: Phoenix.LiveView.Socket.t(), assigns :: map(), action :: atom(), item :: map()) ::
               binary()
 
+  @doc """
+  Customizes the label of the button for creating a new item. Defaults to "New %{resource}".
+  """
+  @callback create_button_label() :: binary()
```
Flo0807, line 127: Should we pass certain parameters to the callback function, e.g. the `assigns`?

mix.exs
```diff
@@ -106,6 +106,7 @@ defmodule Backpex.MixProject do
       "guides/live_resource/panels.md",
       "guides/live_resource/fluid-layout.md",
       "guides/live_resource/listen-to-pubsub-events.md",
+      "guides/live_resource/custom_button_labels.md",
```
pehbehbeh, line 109: We have a lot of short really specific custom guides. Maybe combine with translations?

lib/backpex/live_resource.ex
pehbehbeh: Let's add it when we need it? (YAGNI)

pehbehbeh (approved):

mix.exs
Flo0807: Moved it to translations and renamed translations to "Custom Labels and Translations"

PR #400, opened 2024-06-22
Prevent PETAL from being split at linebreaks

MrMovl:
Minor html improvement to remove linebreaks withing the PETAL acronym.


pehbehbeh:
`<nobr>` is deprecated and should not be used - CSS `white-space` should be preferred.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nobr

MrMovl:
> `<nobr>` is deprecated and should not be used - CSS `white-space` should be preferred.
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nobr

Well damn, weird, I must have believed some older sources that stated something else. I'll fix it

PR #406, opened 2024-06-24
Pass persisted item to `consume_upload` function and pass `data` to item actions and resource actions

Flo0807:
- Pass persisted item with its changes to `consume_upload/4`.
- Pass `data` from `Ecto.Changeset.apply_changes/2`  instead of `params` to item action's `handle` function
- Pass `data` from `Ecto.Changeset.apply_changes/2`  instead of `params` to resource action's `handle` function

### DIFF
--- a/lib/backpex/fields/upload.ex
+++ b/lib/backpex/fields/upload.ex
@@ -54,7 +54,6 @@ defmodule Backpex.Fields.Upload do
 
   **Parameters**
   * `:socket` - The socket.
-  * `:params` (map) - The current params that will be passed to the changeset function.
   * `:item` (struct) - The saved item (with its changes).
   * `:meta` - The upload meta.
   * `:entry` - The upload entry.
--- a/lib/backpex/fields/upload.ex
+++ b/lib/backpex/fields/upload.ex
@@ -55,7 +55,7 @@ defmodule Backpex.Fields.Upload do
   **Parameters**
   * `:socket` - The socket.
   * `:params` (map) - The current params that will be passed to the changeset function.
-  * `:item` (struct) - The item without its changes.
+  * `:item` (struct) - The saved item (with its changes).
   * `:meta` - The upload meta.
   * `:entry` - The upload entry.
 
@@ -63,7 +63,7 @@ defmodule Backpex.Fields.Upload do
 
   **Example**
 
-      defp consume_upload(_socket, _params, _item, %{path: path} = _meta, entry) do
+      defp consume_upload(_socket, _item, %{path: path} = _meta, entry) do
         file_name = ...
         file_url = ...
         static_dir = ...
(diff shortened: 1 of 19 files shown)


lib/backpex/fields/upload.ex
```diff
@@ -55,14 +55,15 @@ defmodule Backpex.Fields.Upload do
   **Parameters**
   * `:socket` - The socket.
   * `:item` (struct) - The item without its changes.
+  * `:change` (map) - The current change / attrs that will be passed to the changeset function.
```
pehbehbeh, line 58: I think we should use only one word here to reduce confusion between change / attrs / params (mentioned elsewhere).

- Are these the form params coming from the user? Then we should call it `params`.
- Are these the changes after casting from the Ecto changeset? Then we should call it `changes` (plural).

https://hexdocs.pm/ecto/Ecto.Changeset.html#module-the-ecto-changeset-struct

Do we have the `changes` available at this point? I think, that would be best (in case we have a choice).

pehbehbeh (changes requested):

lib/backpex/fields/upload.ex
Flo0807: These are the form params. I renamed it to `params`. The form is called `change` in Backpex so this is a little bit confusing. Maybe we should refactor it in Backpex in the future (see [`form_component.ex#L84`](https://github.com/naymspace/backpex/blob/0.4.0/lib/backpex/live_components/form_component.ex#L84))

lib/backpex/live_components/form_component.ex
```diff
@@ -328,7 +328,7 @@ defmodule Backpex.FormComponent do
       %{valid?: true} ->
         result = resource_action.module.handle(socket, params)
 
-        if match?({:ok, _msg}, result), do: handle_uploads(socket)
+        if match?({:ok, _msg}, result), do: handle_uploads(socket, params)
```
Flo0807, line 331: We do not have an item here so what do we want to pass to `handle_uploads`?

Flo0807: We now use `Ecto.Changeset.apply_action/2` and pass the data instead of the params

lib/backpex/resource_action.ex
```diff
@@ -54,11 +54,11 @@ defmodule Backpex.ResourceAction do
             ) :: Ecto.Changeset.t()
 
   @doc """
-  The handle function for the corresponding action. It receives the params and will be called when the form is valid and submitted.
+  The handle function for the corresponding action. It receives the socket and casted and validated data (received from [`Ecto.Changeset.apply_changes/2`](https://hexdocs.pm/ecto/Ecto.Changeset.html#apply_action/2)) and will be called when the form is valid and submitted.
```
pehbehbeh, line 57:
```suggestion
  The handle function for the corresponding action. It receives the socket and casted and validated data (received from [`Ecto.Changeset.apply_changes/1`](https://hexdocs.pm/ecto/Ecto.Changeset.html#apply_action/2)) and will be called when the form is valid and submitted.
```

https://hexdocs.pm/ecto/Ecto.Changeset.html#apply_changes/1

pehbehbeh (approved):

PR #415, opened 2024-06-26
Remove floki

### LINKED ISSUE #402: Remove floki
The Floki dependency is only being used for inserting the Backpex logo (see [`Layout.ex#L245`](https://github.com/naymspace/backpex/blob/0.4.0/lib/backpex/html/layout.ex#L245)).

Maybe we can provide a component for the logo that contains the `svg`. That way we could remove the dependency entirely. 

The dependency causes a conflict in some Phoenix applications because the `only: [:test]` option is set by default in newly generated apps.

related to #399

Flo0807:
closes #402 

- Remove floki
- Update `backpex_logo` component to work without floki


lib/backpex/html/layout.ex
```diff
@@ -230,35 +230,60 @@ defmodule Backpex.HTML.Layout do
...
   """
   @doc type: :component
-  attr(:class, :string,
-    required: false,
-    default: "",
-    doc: "class that will be added to the SVG element"
-  )
+  attr(:class, :string, required: false, default: nil, doc: "class that will be added to the SVG element")
 
   def backpex_logo(assigns) do
```
Flo0807, line 235: I don't know if `layout.ex` is still the right place for the logo component. What's your opinion on that? @pehbehbeh

pehbehbeh: I would remove the component completely and just use `priv/static/images/logo.svg` via `img` tags? Maybe move it to `demo/assets/images` (?) and adjust paths in README and docs.

We do not need to have a component with a dynamic class. We are not changing colors or anything inside the SVG. Size and spacing can be controlled from outside.

pehbehbeh (changes requested):

lib/backpex/html/layout.ex
Flo0807: > Maybe move it to demo/assets/images

We should not move it to the demo directory because then the image is not part of the Backpex package.

We use the logo in Backpex as well as in the demo. If we use `img` tags, we have to copy the SVG into both projects.

The logo is used here:
- [Demo: `index.html.heex#L5`](https://github.com/naymspace/backpex/blob/0.4.0/demo/lib/demo_web/live/home_live/index.html.heex#L5)
- [Demo: `index.html.heex#L114`](https://github.com/naymspace/backpex/blob/0.4.0/demo/lib/demo_web/live/home_live/index.html.heex#L114)
- [Backpex: `layout.ex#L218`](https://github.com/naymspace/backpex/blob/0.4.0/lib/backpex/html/layout.ex#L218)

PR #444, opened 2024-07-03
Move filter_badge component to CoreComponents

makl9797:
Makes the `filter_badge` component reusable for others and move it to the `Backpex.HTML.CoreComponents` module.


lib/backpex/html/core_components.ex
```diff
@@ -13,4 +13,35 @@ defmodule Backpex.HTML.CoreComponents do
     <span class={[@name, @class]} {@rest} />
     """
   end
+
+  @doc """
+  Renders a filter_badge component.
+  """
+  attr :clear_event, :string, doc: "event name for removing the badge"
```
Flo0807, line 20: We should add a default here:

```suggestion
  attr :clear_event, :string, default: "clear-filter", doc: "event name for removing the badge"
```

Flo0807 (changes requested):

lib/backpex/html/core_components.ex
makl9797: Good idea 👍

PR #457, opened 2024-07-10
Fix `update_item` may cause errors

Flo0807:
It's possible that `update_item` raises an `Ecto.NoResultsError`. For example, if the `item_query` adds additional filters (e.g., limit access to specific users). We would rather not raise in this case.

### DIFF
--- a/lib/backpex/resource.ex
+++ b/lib/backpex/resource.ex
@@ -254,7 +254,7 @@ defmodule Backpex.Resource do
   end
 
   @doc """
-  Gets a database record with the given fields by the given id.
+  Gets a database record with the given fields by the given id. Raises Ecto.NoResultsError if no record was found.
 
   ## Parameters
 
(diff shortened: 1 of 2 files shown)


lib/backpex/resource.ex
```diff
@@ -254,7 +254,7 @@ defmodule Backpex.Resource do
   end
 
   @doc """
-  Gets a database record with the given fields by the given id.
+  Gets a database record with the given fields by the given id. Raises Ecto.NoResultsError if no record was found.
```
pehbehbeh, line 257:
```suggestion
  Gets a database record with the given fields by the given id. Raises `Ecto.NoResultsError` if no record was found.
```

lib/backpex/resource.ex
```diff
@@ -265,6 +265,25 @@ defmodule Backpex.Resource do
...
+  ## Parameters
+
+  * `id`: The identifier for the specific item to be fetched.
+  * `repo` (module): The repository module.
+  * `schema`: The Ecto schema module corresponding to the item
+  * `item_query` (function): A function that modifies the base query. This function should accept an Ecto.Queryable and return an Ecto.Queryable. It's used to apply additional query logic.
+  * `fields` (list): A list of atoms representing the fields to be selected and potentially preloaded.
+  """
+  def get(id, repo, schema, item_query, fields) do
+    query_record(id, repo, schema, item_query, fields, :one)
```
pehbehbeh, line 283: I find that a bit difficult to read. Maybe write it like that?

```suggestion
    record_query(id, repo, schema, item_query, fields)
    |> Repo.one!()
  end

  @doc """
  Gets a database record with the given fields by the given id. Returns nil if no result was found.

  ## Parameters

  * `id`: The identifier for the specific item to be fetched.
  * `repo` (module): The repository module.
  * `schema`: The Ecto schema module corresponding to the item
  * `item_query` (function): A function that modifies the base query. This function should accept an Ecto.Queryable and return an Ecto.Queryable. It's used to apply additional query logic.
  * `fields` (list): A list of atoms representing the fields to be selected and potentially preloaded.
  """
  def get(id, repo, schema, item_query, fields) do
    record_query(id, repo, schema, item_query, fields)
    |> Repo.one()
```

PR #468, opened 2024-07-13
daisyUI theme selector

thomasfortes:
This is a theme selector similar to the one used in the daisyui site, I'm looking for some opinions.

First of all in the `app_shell` I refactored the `topbar` rendering to be called only once, it was needed because I need an id to get the form value to change the theme using js and the double rendering is causing multiple ids errors, so if there's any issues let me know.

Then there's also repeated code of `defp cookie_path`, `theme_selector` is a layout element and I just copied the private function inside `resource.ex`, let me know if there's a problem with that.

How it works:
Whenever someone clicks in a theme I dispatch an event that is listened by a small js snippet that gets the selected element, update the `data-theme` attribute in the `html` element  and make an async request to the cookie_path to save the selected theme in the session.

This works _mostly_ fine but there's an issue when you select a theme and then close the tab and reopen quickly (with a keyboard shortcut) without navigating to any other route, sometimes the browser reuse the session and render the old theme, I tested a solution that stores the theme in localstorage and restore when the app.js is loaded but didn't commit because I want some feedback if that's something you're okay with.

After we figure out the code issues I can work the docs.


lib/backpex/html/layout.ex
```diff
@@ -226,6 +220,68 @@ defmodule Backpex.HTML.Layout do
...
+        <Backpex.HTML.CoreComponents.icon name="hero-swatch-solid" class="h-5 w-5 md:hidden" />
+        <svg
+          width="12px"
+          height="12px"
+          class="inline-block h-2 w-2 fill-current opacity-60"
+          xmlns="http://www.w3.org/2000/svg"
+          viewBox="0 0 2048 2048"
+        >
+          <path d="M1799 349l242 241-1017 1017L7 590l242-241 775 775 775-775z"></path>
+        </svg>
```
Flo0807, line 250: We should use a Heroicon. Maybe `chevron-down`?

lib/backpex/html/layout.ex
```diff
@@ -226,6 +220,68 @@ defmodule Backpex.HTML.Layout do
...
+  )
+
+  def theme_selector(assigns) do
+    ~H"""
+    <div id="backpex-theme-selector" class="dropdown dropdown-bottom dropdown-end no-animation">
+      <div tabindex="0" role="button" class="btn m-1">
+        <span class="hidden md:block">
+          <%= Backpex.translate("Theme") %>
+        </span>
+        <Backpex.HTML.CoreComponents.icon name="hero-swatch-solid" class="h-5 w-5 md:hidden" />
```
Flo0807, line 241: I vote for the outline version of the `swatch` icon because the user icon is also outline.

Flo0807:
> First of all in the app_shell I refactored the topbar rendering to be called only once, it was needed because I need an id to get the form value to change the theme using js and the double rendering is causing multiple ids errors, so if there's any issues let me know.

I don't think this is an issue. I have tried it and everything seems to work fine on both mobile and desktop.

> Then there's also repeated code of defp cookie_path, theme_selector is a layout element and I just copied the private function inside resource.ex, let me know if there's a problem with that.

Maybe we can move the `cookie_path` function to `router.ex`. What do you think?

I have also noticed that the theme is not selected in the selector after refreshing the page.

lib/backpex/html/layout.ex
thomasfortes: Agree, also using btn-ghost to keep in line with the user icon.

thomasfortes:
> Maybe we can move the `cookie_path` function to `router.ex`. What do you think?

Perfect, let me know if you want something different but I already moved it to the Router

> I have also noticed that the theme is not selected in the selector after refreshing the page.

Also fixed, the only "issue" is that if you're using a theme from the bottom of the list it will not be in focus when you open the dropdown, but that is a limitation daisyui site also has.

Also, don't know what themes do you want for the demo site, right now I'm loading every official theme.

Flo0807:
> Also, don't know what themes do you want for the demo site, right now I'm loading every official theme.

Every theme is fine, I think 👍

Flo0807:
Hey @thomasfortes!

This PR is still a draft. I'd love to see a theme switch in the demo. Do you need help or have any questions?

thomasfortes:
I think that everything is working right now, let me know if you have any concerns

Flo0807:
@thomasfortes I finally got back to this PR. I made some adjustments to the installation guide and merged the current `develop` branch.

Thanks for your work. I really like how the theme switch turned out! 🚀

PR #478, opened 2024-07-19
Add support for serial ids in HasMany field

### LINKED ISSUE #451: Process crash when selecting a value in ManyToMany field
Hi,

First thank you for this amazing library 🙏 🙏 🙏 

I'm having an issue creating a `ManyToMany` field.
Whenever I try to select a value (in this case a `Label` for a `Sample`) the process crashes.

Here is the trace:
```
[error] GenServer #PID<0.4326.0> terminating
** (BadMapError) expected a map, got: nil
    (elixir 1.17.1) lib/map.ex:535: Map.get(nil, :name, nil)
    (backpex 0.5.1) lib/backpex/fields/many_to_many.ex:386: Backpex.Fields.ManyToMany.to_form_option/2
    (elixir 1.17.1) lib/enum.ex:1703: Enum."-map/2-lists^map/1-1-"/2
    (backpex 0.5.1) lib/backpex/fields/many_to_many.ex:210: Backpex.Fields.ManyToMany.render_form/1
    (phoenix_live_view 1.0.0-rc.6) lib/phoenix_live_view/renderer.ex:79: Phoenix.LiveView.Renderer.to_rendered/2
    (phoenix_live_view 1.0.0-rc.6) lib/phoenix_live_view/diff.ex:335: Phoenix.LiveView.Diff.component_to_rendered/3
    (phoenix_live_view 1.0.0-rc.6) lib/phoenix_live_view/diff.ex:753: Phoenix.LiveView.Diff.render_component/8
    (phoenix_live_view 1.0.0-rc.6) lib/phoenix_live_view/diff.ex:214: Phoenix.LiveView.Diff.write_component/4
    (phoenix_live_view 1.0.0-rc.6) lib/phoenix_live_view/channel.ex:662: Phoenix.LiveView.Channel.component_handle/4
    (stdlib 6.0) gen_server.erl:2173: :gen_server.try_handle_info/3
    (stdlib 6.0) gen_server.erl:2261: :gen_server.handle_msg/6
    (stdlib 6.0) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
```

I'd appreciate your help in finding what am I doing wrong.

❤️

thomasfortes:
Add a match function that handles when `option.id` is represented as an integer in the HasMany field.

Closes #451 ?

### DIFF
--- a/lib/backpex/fields/has_many.ex
+++ b/lib/backpex/fields/has_many.ex
@@ -244,15 +244,15 @@ defmodule Backpex.Fields.HasMany do
       }
     } = socket
 
-    selected_item = Enum.find(selected, fn option -> option.id == id end)
+    selected_item = Enum.find(selected, fn option -> match_option_to_id(option, id) end)
 
     new_selected =
       if selected_item do
-        Enum.reject(selected, fn option -> option.id == id end)
+        Enum.reject(selected, fn option -> match_option_to_id(option, id) end)
       else
         selected
         |> Enum.reverse()
-        |> Kernel.then(&[Enum.find(options, fn option -> option.id == id end) | &1])
+        |> Kernel.then(&[Enum.find(options, fn option -> match_option_to_id(option, id) end) | &1])
         |> Enum.reverse()
       end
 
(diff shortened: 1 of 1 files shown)


Flo0807:
> The only thing that I can think of is that when the internal representation is guarded to be an integer and String.to_integer fails it probably means that something else is wrong and knowing it earlier is better than letting it fail silently.

True! Let's try it with your solution 👍

PR #481, opened 2024-07-19
Fix `Backpex.Resource.update_all/6` when used without pubsub arg

asok:
Before this change when someone called the function with 5 arguments `FunctionClauseError` was risen.

### DIFF
--- a/lib/backpex/resource.ex
+++ b/lib/backpex/resource.ex
@@ -503,9 +503,7 @@ defmodule Backpex.Resource do
     {:ok, item}
   end
 
-  defp broadcast({:ok, item} = event, _event, nil) do: event
-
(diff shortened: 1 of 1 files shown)


lib/backpex/resource.ex
```diff
@@ -503,6 +503,8 @@ defmodule Backpex.Resource do
     {:ok, item}
   end
 
+  defp broadcast({:ok, item} = event, _event, nil) do: event
+
   defp broadcast({:error, _reason} = event, _event, _pubsub), do: event
```
Flo0807, line 508:
```suggestion
  defp broadcast(result, _event, _pubsub), do: result
```

We can handle both cases with one function clause.

Flo0807 (changes requested):

Flo0807:
Thanks 🙂

PR #490, opened 2024-07-26
Update installation.md

mradke:
There was a missing `end` delimiter in the example which might be hard to spot for new users.


Flo0807:
Thanks for the fix!

PR #492, opened 2024-07-26
Add parentheses to functions

Flo0807:


Flo0807:
See https://github.com/naymspace/backpex/pull/513

PR #494, opened 2024-07-28
LiveResource generator

phyr97:
I tried to work with some code generators and wanted to implement a simple generator for a Backpex.LiveResource. This generator serves as a starting point to take an existing schema module and create a Backpex.LiveResource. Simple field types like :integer, :string, and :boolean are converted into their corresponding Backpex types.

I believe there are many possibilities to enhance this generator or create additional ones, such as for item actions or resource actions.

This is my first time creating a generator like this, so there may be some aspects that need to be handled differently.

### DIFF
--- a/lib/mix/backpex/live_resource.ex
+++ b/lib/mix/backpex/live_resource.ex
@@ -15,6 +15,13 @@ defmodule Mix.Backpex.LiveResource do
             plural_name: nil,
             fields: []
 
+  @switches [
+    integer: Backpex.Fields.Number,
+    boolean: Backpex.Fields.Boolean,
+    utc_datetime: Backpex.Fields.DateTime,
+    naive_datetime: Backpex.Fields.DateTime,
+    string: Backpex.Fields.Text
+  ]
   def new(schema_name, opts) do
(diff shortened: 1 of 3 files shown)


Flo0807:
LiveResource generator:

**PubSub**

```elixir
# config.exs
config :backpex, :pubsub_server, MyApp.PubSub
```

**Backpex Hooks**

**daisyUI**

`npm i -D daisyui@latest`

Add to `app.css`:
```css
# app.css
@plugin "daisyui";
```

**Setup formatter**

```elixir
# my_app/.formatter.exs
[
  import_deps: [:backpex]
]
```

**Create Layout** (may be disabled via flag)


**Add Backpex Routes**

```elixir
# router.ex

import Backpex.Router

scope "/", MyAppWeb do
   backpex_routes()
end
```

add live session and `/admin` path?

**Add hint for `tailwindcss/forms` plugin**

If forms plugin is present in`app.css` ask to remove it because of the conflict between daisyUI and the plugin

**Add hint for body background color**

Search for `body class="bg-white`. If present output warning.

phyr97:
I'm closing this PR because it will get a fresh start with Igniter in the future.

PR #500, opened 2024-07-31
Fix item action guide

krns:


guides/actions/item-actions.md
```diff
@@ -151,6 +153,9 @@ defmodule DemoWeb.ItemAction.SoftDelete do
     @impl Backpex.ItemAction
     def label(_assigns), do: Backpex.translate("Delete")
 
+    @impl Backpex.ItemAction
+    def confirm(_assigns), do: "Why do you want to delete this item?"
```
krns, line 157: The confirm function is mandatory.

krns: We could omit the `confirm_label` and `cancel_label`

Flo0807: The `confirm/1` function determines whether a confirmation modal is shown to the user. We can decide to add the `confirm/1`, `confirm_label/1` and `cancel_label/1` functions (a modal is shown) or remove them (no modal is shown) 🤔.

If we add the `confirm/1` function as suggested, we must also keep `confirm_label/1` and `cancel_label/1`.

krns: The current example in the documentation makes no sense (`confirm_label/1` and `cancel_label/1` without modal).

> If we add the confirm/1 function as suggested, we must also keep `confirm_label/1` and `cancel_label/1`.

Thats not right. When omitted the default values ("Cancel" and "Apply") are used.

Flo0807: I did not see the default implementation 🙃

Flo0807:
Currently, the item action confirmation dialog is enabled when the list of fields is empty or the `confirm/1` function is exported (see [`has_modal?/1`](https://github.com/naymspace/backpex/blob/0.6.0/lib/backpex/html/resource.ex#L936)).

From our docs:
> To enable the confirmation dialog you need to implement the `confirm_label/1` function 

This doesn't seem right and has to be updated. 

Moreover, we should update the `has_modal?/1` function because you always need the `confirm/1` function if there is a list of fields.

@krns

PR #529, opened 2024-08-19
Fix HasMany and MultiSelect search field not working

thomasfortes:
Fix #520


Flo0807:
Thanks for the fix 🙏 I also added it to the MultiSelect field. It was broken there too.

PR #539, opened 2024-08-23
Fix misplaced bracket in installation instructions

jamescarr:
Noted during setup that lsp was complaining in the config about misplaced closing brackets, so made a quick fix here.


Flo0807:
🚀

PR #541, opened 2024-08-26
Make formatting for number fields optional

makl9797:
This PR removes the formatting for a value of a number field. This should lead to the behavior that formatting is optional and is not carried out by default. 

Formatting of the number field must now be defined with the `render` option on the field itself. If number fields have benefited from the previous standard formatting, the following `render` function can now be used as an example:

```elixir
render: fn assigns ->
  ~H"""
  <p><%= Number.Delimit.number_to_delimited(@value, precision: 0, delimiter: ".") %></p>
  """
end
```

### DIFF
--- a/lib/backpex/html/html.ex
+++ b/lib/backpex/html/html.ex
@@ -15,7 +15,7 @@ defmodule Backpex.HTML do
       "—"
 
       iex> Backpex.HTML.pretty_value(1_000_000)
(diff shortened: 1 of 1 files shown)


Flo0807:
I changed the title of this PR because the term "number field" can be misleading, as formatting was performed for all Backpex fields.

PR #554, opened 2024-08-30
Refactor HasMany

Flo0807:


lib/backpex/fields/has_many.ex
```diff
@@ -293,40 +231,71 @@ defmodule Backpex.Fields.HasMany do
...
-        socket
-        |> assign(:options, options)
-        |> assign(:selected, options)
-        |> assign(:show_more, false)
-      else
-        options = options(assigns, limit: query_limit(field_options))
+  @impl Backpex.Field
+  def before_changeset(changeset, attrs, _metadata, repo, field, assigns) do
+    {field_name, field_options} = field
+    validate_live_resource(field_name, field_options)
```
pehbehbeh, line 249: Things like this should be done way more generic and for every field / filter / etc. (in the future).

https://github.com/dashbitco/nimble_options

Maybe even on compile time.

demo/lib/demo/user.ex
```diff
@@ -81,4 +82,14 @@ defmodule Demo.User do
...
+
+  defp validate_posts(changeset) do
+    posts = get_field(changeset, :posts) || []
+
+    if length(posts) < 1 do
+      add_error(changeset, :posts, "must have at least one post")
+    else
+      changeset
+    end
+  end
```
pehbehbeh, line 94: Just an idea: We could provide a generic version of this via `Backpex.Ecto.ChangesetHelpers.validate_assoc_length` or something like that (in the future).

lib/backpex/fields/has_many.ex
```diff
@@ -381,13 +332,155 @@ defmodule Backpex.Fields.HasMany do
...
+    selected_from_db =
+      if Enum.empty?(ids_to_fetch) do
+        []
+      else
+        queryable
+        |> where([x], x.id in ^ids_to_fetch)
+        |> maybe_options_query(field_options, assigns)
+        |> select([x], {field(x, ^display_field_form(field)), x.id})
+        |> repo.all()
+      end
```
pehbehbeh, line 451: Should be its own functions to improve readability here.

You also don't need the big pattern matching block at the start of the function when splitting things up. It is also OK to write `assigns.foo` in some places, I think.

pehbehbeh (approved):

PR #556, opened 2024-08-30
Fixed index_editable in Backpex.Fields.Number

nedhgx:
Fixed index_editable in Backpex.Fields.Number

When `index_editable: true` is set in the field with Backpex.Fields.Number, it will fail to update.


Flo0807:
🚀🚀

PR #588, opened 2024-09-06
Remove unused dependencies

Flo0807:
Removed from `package.json`:
- `lodash`

Removed from `mix.exs`:
- `swoosh`
- `phoenix_swoosh`
- `gen_smtp`
- `hackney`

Also cleaned `phoenix_view` and `ranch` from `mix.lock`


demo/mix.exs
```diff
@@ -57,9 +57,6 @@ defmodule Demo.MixProject do
       {:ex_machina, "~> 2.3"},
       {:hackney, "~> 1.17", override: true},
       {:faker, "~> 0.18"},
-      {:swoosh, "~> 1.0"},
-      {:phoenix_swoosh, "~> 1.0"},
-      {:gen_smtp, "~> 1.1"},
```
Flo0807, line 62: Are there any other deps we do not use? 🤔

pehbehbeh: Maybe `tesla` and `hackney`?

pehbehbeh (approved):

demo/mix.exs
Flo0807: `hackney` is required by Sentry, but I removed `tesla`

PR #610, opened 2024-09-23
Switch from Alpine to Ubuntu

Flo0807:
- Remove bashrc stuff
- Switch from alpine to ubuntu
- Configure `renovate.json` to properly update elixir base image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,9 +33,6 @@ RUN curl -fsSL https://deb.nodesource.com/setup_lts.x | bash - \
     && apt-get install -y nodejs \
     && npm install --global yarn
 
-
-SHELL ["/bin/bash", "-c"]
-
 COPY .docker/opt/scripts/ /opt/scripts
 ADD https://github.com/naymspace/env-secrets-expand/raw/main/env-secrets-expand.sh /opt/scripts/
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -71,23 +83,18 @@ RUN mix do deps.get, deps.compile, assets.deploy, sentry.package_source_code, re
...
+    && apt-get install -y libstdc++6 openssl libncurses5 locales ca-certificates wget \
+    && apt-get clean && rm -f /var/lib/apt/lists/*_*
 
-COPY --from=builder /etc/bashrc.d /etc/bashrc.d
-COPY --from=builder /etc/bash.bashrc /etc/bash.bashrc
 COPY --from=builder /opt/scripts /opt/scripts
-COPY .docker/root/.bashrc /root/
 
-RUN chown -R nobody:nobody /opt
+RUN chown -R nobody /opt
```
pehbehbeh, line 97: This needs to be done in the two places below (`COPY`, `USER`), too - I think.

pehbehbeh (changes requested):

Dockerfile
Flo0807: 👍

PR #635, opened 2024-10-17
Add save & continue editing button

### LINKED ISSUE #330: Save & continue editing
Between Cancel and Save, there should be a "Save & continue editing" button. By clicking the button, you can save the form without getting redirected to the index / show view.

Elmseld:
closes #330


lib/backpex/html/resource/form_component.html.heex
```diff
@@ -53,7 +53,19 @@
           </button>
         </.link>
         <button
+          :if={@continue_label}
           type="submit"
+          name="save-type"
+          value="continue"
+          class={["btn", if(form_errors?(@show_form_errors, @form), do: "btn-error", else: "btn-primary")]}
```
pehbehbeh, line 60: Maybe this should be `btn-neutral btn-outline` to not have two primary buttons at the same time? The primary button should indicate "enter" press.

lib/backpex/live_components/form_component.ex
```diff
@@ -309,7 +313,7 @@ defmodule Backpex.FormComponent do
     end
   end
 
-  defp handle_save(socket, :resource_action, params) do
+  defp handle_save(socket, :resource_action, params, _) do
```
pehbehbeh, line 316: Please add parameter names.

btw... There should also be a Credo rule for that. Maybe you can enable it (if its not to much work in the rest of the codebase)?

lib/backpex/live_components/form_component.ex
```diff
@@ -422,6 +426,22 @@ defmodule Backpex.FormComponent do
...
 
+  defp return_to_path("continue", _, _, %{current_url: url}, item, :new) do
+    url
+    |> URI.parse()
+    |> Map.get(:path)
+    |> Path.dirname()
+    |> Kernel.<>("/#{item.id}/edit")
+  end
+
+  defp return_to_path("continue", _, _, %{current_url: url}, _, _) do
```
pehbehbeh, line 437: add `:edit` here to have an explicit case

lib/backpex/live_components/form_component.ex
```diff
@@ -422,6 +426,22 @@ defmodule Backpex.FormComponent do
...
+    URI.parse(url).path
+  end
+
+  defp return_to_path(_, live_resource, socket, assigns, item, type) do
```
pehbehbeh, line 441: add `"save"` here to have an explicit case

pehbehbeh (changes requested):

lib/backpex/html/layout.ex
```diff
@@ -130,6 +130,12 @@ defmodule Backpex.HTML.Layout do
       :if={Phoenix.Flash.get(@flash, :info) && Phoenix.Flash.get(@flash, :info) != ""}
       class="alert bg-info text-info-content my-4 text-sm"
       phx-value-key="info"
+      x-data="{ isVisible: false }"
+      x-init="() => { isVisible = true; setTimeout(() => isVisible = false, 5000) }"
+      x-show="isVisible"
+      x-transition:leave="transition ease-in duration-2000"
+      x-transition:leave-start="opacity-100 transform scale-100"
+      x-transition:leave-end="opacity-0 transform scale-90"
```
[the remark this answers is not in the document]
Elmseld: The best I can do is either this: 

```
    <script>
      setTimeout(() => {
        const infoFlash = document.getElementById("info-flash");

        if (infoFlash) {
          infoFlash.style.transition = "opacity 2s ease-in";
          infoFlash.style.opacity = 0;
          setTimeout(() => infoFlash.remove(), 2000);
        }
      }, 5000);
    </script>
```
in the html in the flash_messages function(not pretty), since there is no assets with js I can't create a hook.

Or something like this: 

```
    <div
      :if={Phoenix.Flash.get(@flash, :info) && Phoenix.Flash.get(@flash, :info) != ""}
      class="alert bg-info text-info-content my-4 text-sm"
      id="info-flash"
      phx-click-away={JS.hide(to: "#info-flash", transition: {"transition ease-out duration-2000", "opacity-100 transform scale-100", "opacity-0 transform scale-90"})}
    >
```

That removes the msg when a user is using the page again. But that will remove the msg immediately on click-away with might make it disappear to fast.

.credo.exs
```diff
@@ -16,7 +16,10 @@
       parse_timeout: 5000,
       color: true,
       checks: %{
-        disabled: []
+        disabled: [],
+        enabled: [
+          {Credo.Check.Consistency.UnusedVariableNames, []}
+        ]
```
Flo0807, line 22: We cannot do this as this will result in only adding the `Credo.Check.Consistency.UnusedVariableNames` check.

> ❯ mix credo
Checking 51 source files ...
Analysis took 0.06 seconds (0.06s to load, 0.00s running 1 check on 51 files)

Note that the log displays `1 check` only.

The reason is that credo automatically adds all default checks if the `enabled` key is not defined. If its there, it will use the configured checks only.

Instead of:

```elixir
checks: %{
  disabled: [],
  enabled: [
    {Credo.Check.Consistency.UnusedVariableNames, []}
  ]
}
```

we can do this

```elixir
checks: [
  {Credo.Check.Consistency.UnusedVariableNames, []}
]
```

Elmseld: I updated this so now there is: `Analysis took 1.8 seconds (0.05s to load, 1.7s running 67 checks on 52 files)`

Flo0807 (approved):

PR #645, opened 2024-10-25
Automerge github actions dependency updates

Flo0807:


Flo0807:
Since we are pinning GitHub actions dependencies and therefore updates happen very frequently, I think this will reduce some maintenance work for you @krns

PR #650, opened 2024-11-03
Improve ash setup

Flo0807:
- Reuse `Demo.Repo` for Ash
- Add `min_pg_version` function required by #647 
- Add spark formatter
- Use `AshPostgres.Repo` instead of `Ecto.Repo`


demo/lib/demo/repo.ex
```diff
@@ -2,12 +2,12 @@ defmodule Demo.Repo do
   use Ecto.Repo,
```
pehbehbeh, line 2: If we combine the two: Should we use `AshPostgres.Repo` then?

> This repo is a thin wrapper around an `Ecto.Repo`.

https://hexdocs.pm/ash_postgres/AshPostgres.Repo.html

Then we also would not have to implement `min_pg_version`.

Flo0807: yes, then we should switch!

Flo0807: done

pehbehbeh (approved):

demo/lib/demo/repo.ex
Flo0807: @pehbehbeh you should implement `min_pg_version/0`. Otherwise, there is a warning. See https://github.com/ash-project/ash_postgres/blob/v2.4.11/lib/repo/before_compile.ex#L8 and https://github.com/naymspace/backpex/actions/runs/11704207105/job/32596160978

PR #651, opened 2024-11-03
improvement: don't pin AshPostgres to an old version

zachdaniel:
Most users will not be able to use Ash + Backpex with this postgres version requirement as it pins to a very old version of AshPostgres.


pehbehbeh:
Good catch, thanks!

PR #652, opened 2024-11-04
Add `Backpex.Fields.Time`

thanos:
Although it seems a trivial field to add and seemingly redundant due to the DateTime field, Time on its own used  in finance. 
Typically examples are when you need to represent cutoff SLA, such as this feed must arrive by 12:35 each day. Obviously a text field representing a CRON works (working on it) but 99.999%  of the business would refuse to understand. So since Ecto has Time why not Backpack? 

BTW many thanks for this amazing library!


Flo0807:
Hey @thanos, thanks for the contribution. I have updated the field to the latest version of Backpex.

You will be able to use the field with the next release 🚀

PR #655, opened 2024-11-07
improve credo config

pehbehbeh:
- enable all credo checks by default
- overwrite (disable or reconfigure) credo checks explicitly
- implement some easy credo checks and fixes

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -1,3 +1,29 @@
+overwrite_checks = [
+  {Credo.Check.Consistency.ExceptionNames, false},
+  {Credo.Check.Design.TagTODO, false},
+  {Credo.Check.Readability.AliasAs, false},
+  {Credo.Check.Readability.OnePipePerLine, false},
+  {Credo.Check.Readability.SinglePipe, false},
+  {Credo.Check.Readability.Specs, false},
+  {Credo.Check.Readability.StrictModuleLayout, ignore_module_attributes: ~w[config_schema]a},
+  {Credo.Check.Refactor.ABCSize, false},
+  {Credo.Check.Refactor.CyclomaticComplexity, false},
+  {Credo.Check.Refactor.ModuleDependencies, false},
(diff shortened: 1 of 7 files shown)


.credo.exs
```diff
@@ -1,3 +1,27 @@
...
+  {Credo.Check.Readability.OnePipePerLine, false},
+  {Credo.Check.Readability.SinglePipe, false},
+  {Credo.Check.Readability.Specs, false},
+  {Credo.Check.Readability.StrictModuleLayout, ignore_module_attributes: ~w[config_schema]a},
+  {Credo.Check.Refactor.ABCSize, false},
+  {Credo.Check.Refactor.ModuleDependencies, false},
+  {Credo.Check.Refactor.PipeChainStart, false},
+  {Credo.Check.Refactor.VariableRebinding, false},
+  {Credo.Check.Warning.LazyLogging, false}
+]
```
Flo0807, line 13: I'd enable `Credo.Check.Readability.SinglePipe` and maybe also `Credo.Check.Warning.LazyLogging`

Flo0807 (approved):

.credo.exs
pehbehbeh: `LazyLogging` is not supported on Elixir version 1.7 or greater.

```
info: some checks were skipped because they're not compatible with
your version of Elixir (1.17.3).

You can deactivate these checks by adding this to the `checks` list in your config:

    {Credo.Check.Warning.LazyLogging, false},
```

https://hexdocs.pm/credo/Credo.Check.Warning.LazyLogging.html

pehbehbeh: `SinglePipe` is in conflict with `NestedFunctionCalls` and I often find it easier to read if there are multiple lines. There are also 90 occurrences (84 backpex, 6 demo) to fix if we enable it.

My goal would be to enable even more rules (like `ABCSize` and `ModuleDependencies`), but these require extensive refactoring and should be its own PRs.

PR #658, opened 2024-11-08
Refactor `field_input` component

### LINKED ISSUE #354: Reusable `field_input` component
We should make the `field_input` component generic so that it can be used by other projects as well as wherever we need form inputs in backpex (e.g., filter).

Flo0807:
closes #354 

I made `field_input` component more generic by not relying on a `field_options` map anymore. I also renamed the component to `input`.

- Rename `field_input` component to `input`
- Remove `field_options` attribute from `input` component
- Add `translate_error_fun` attribute for custom error translations
- Change type `string` to `any` for class attributes passed to `input` component (mainly to include lists, too)
- Refactor all `field_input` calls

### DIFF
--- a/guides/fields/custom-fields.md
+++ b/guides/fields/custom-fields.md
@@ -32,7 +32,7 @@ def render_form(assigns) do
     <:label>
         <Layout.input_label text={@field_options[:label]} />
     </:label>
-    <BackpexForm.field_input
+    <BackpexForm.input
         type="text"
         field={@form[@name]}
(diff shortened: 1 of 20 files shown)


Flo0807:
I also searched for other inputs, e.g., used for the filters. But atm there are no advantages by using the `input` component there in my opinion. For example, the range input is very custom because it displays an additional label inside the form ("From" / "To") @pehbehbeh

PR #659, opened 2024-11-08
Refactor Long Quote Block in LiveResource

pehbehbeh:
see https://hexdocs.pm/credo/Credo.Check.Refactor.LongQuoteBlocks.html

- move everything from quote block to top level module in `Backpex.LiveResource`
- change some internal function to private; see f7f124b3edaa2e657ef7d296d158ac8c75878949
- remove `Backpex.LiveResource.schema/0` function; see 709bff8dac65ec135d6328d79fcd1c7ffb3ab8ef
- fix live updates for non-id primary keys; see dc051d2815bd364b32e43c2c296d4ee37eb03157
- remove obsolete `Backpex.LiveResource.can?/4` helper; see 9bdd735533a69ea51401bc7c2b8282acca949e7d and a4742105a06155e8d8da84d7084ff817ca2b5f16
- simplify active filters function; see 2aad9baa0d708e9d6bf93d3d7d9c09739dfcd7cd
- simplify `get_filter_options` function; see 2a43e79ea9e082272c731b592c739838452b4321

### DIFF
--- a/lib/backpex/fields/has_many.ex
+++ b/lib/backpex/fields/has_many.ex
@@ -335,13 +335,12 @@ defmodule Backpex.Fields.HasMany do
       socket: socket,
       field_options: field_options,
       item: item,
-      live_resource: live_resource,
       params: params,
       link_assocs: link_assocs
     } = assigns
 
     link =
-      if link_assocs and Map.has_key?(field_options, :live_resource) and live_resource.can?(assigns, :show, item) do
+      if link_assocs and field_options.live_resource.can?(assigns, :show, item) do
         Router.get_path(socket, Map.get(field_options, :live_resource), params, :show, item)
       else
         nil
--- a/lib/backpex/fields/has_many.ex
+++ b/lib/backpex/fields/has_many.ex
@@ -39,7 +39,6 @@ defmodule Backpex.Fields.HasMany do
   import Ecto.Query
   import Backpex.HTML.Form
   alias Backpex.Adapters.Ecto, as: EctoAdapter
-  alias Backpex.LiveResource
   alias Backpex.Router
 
   @impl Phoenix.LiveComponent
@@ -341,8 +340,7 @@ defmodule Backpex.Fields.HasMany do
     } = assigns
 
     link =
-      if link_assocs and Map.has_key?(field_options, :live_resource) and
-           LiveResource.can?(assigns, :show, item, live_resource) do
+      if link_assocs and Map.has_key?(field_options, :live_resource) and live_resource.can?(assigns, :show, item) do
         Router.get_path(socket, Map.get(field_options, :live_resource), params, :show, item)
       else
         nil
--- a/lib/backpex/fields/has_many.ex
+++ b/lib/backpex/fields/has_many.ex
@@ -255,7 +255,8 @@ defmodule Backpex.Fields.HasMany do
     {field_name, field_options} = field
     validate_live_resource(field_name, field_options)
 
-    schema = field_options.live_resource.schema()
+    # TODO: do not rely on specific adapter
+    schema = field_options.live_resource.config(:adapter_config)[:schema]
     field_name_string = to_string(field_name)
 
     new_assocs = get_new_assocs(attrs, field_name_string, schema, repo, field_options, assigns)
--- a/lib/backpex/live_resource.ex
+++ b/lib/backpex/live_resource.ex
@@ -1182,18 +1182,20 @@ defmodule Backpex.LiveResource do
     |> maybe_assign_metrics()
   end
 
-  defp update_item(socket, %{id: id} = _item) do
+  defp update_item(socket, item) do
     %{
       live_resource: live_resource,
       live_action: live_action
     } = socket.assigns
 
-    item = Resource.get(id, socket.assigns, live_resource)
+    item_primary_value = primary_value(socket, item)
+    item = Resource.get(item_primary_value, socket.assigns, live_resource)
 
     socket =
       cond do
         live_action in [:index, :resource_action] and item ->
-          items = Enum.map(socket.assigns.items, &if(primary_value(socket, &1) == id, do: item, else: &1))
+          items =
+            Enum.map(socket.assigns.items, &if(primary_value(socket, &1) == item_primary_value, do: item, else: &1))
 
           assign(socket, :items, items)
 
--- a/lib/backpex/live_resource.ex
+++ b/lib/backpex/live_resource.ex
@@ -221,11 +221,6 @@ defmodule Backpex.LiveResource do
   @callback resource_created_message() :: binary()
 
   @doc """
-  Returns the schema of the live resource.
-  """
-  @callback schema() :: module()
-
-  @doc """
   Uses LiveResource in the current module to make it a LiveResource.
 
       use Backpex.LiveResource,
@@ -300,9 +295,6 @@ defmodule Backpex.LiveResource do
 
       @impl Backpex.LiveResource
       def create_button_label, do: Backpex.translate({"New %{resource}", %{resource: singular_name()}})
-
-      @impl Backpex.LiveResource
-      def schema, do: @resource_opts[:adapter_config][:schema]
 
       @impl Backpex.LiveResource
       def resource_created_message,
--- a/lib/backpex/live_resource.ex
+++ b/lib/backpex/live_resource.ex
@@ -514,22 +514,14 @@ defmodule Backpex.LiveResource do
 
   defp field_active?(name, saved_fields) do
     case Map.get(saved_fields, Atom.to_string(name)) do
-      "true" ->
-        true
-
-      "false" ->
-        false
-
-      _other ->
-        true
+      "true" -> true
+      "false" -> false
+      _other -> true
     end
   end
 
   def assign_items(socket) do
-    %{
-      live_resource: live_resource,
-      fields: fields
-    } = socket.assigns
+    %{live_resource: live_resource, fields: fields} = socket.assigns
 
     criteria = build_criteria(socket.assigns)
     items = Resource.list(fields, socket.assigns, live_resource, criteria)
@@ -586,7 +578,7 @@ defmodule Backpex.LiveResource do
   end
 
   @impl Phoenix.LiveView
-  def render(%{live_action: action} = assigns) when action in [:show, :show_edit] do
+  def render(%{live_action: action} = assigns) when action in [:show] do
     resource_show(assigns)
   end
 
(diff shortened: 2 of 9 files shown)


pehbehbeh:
@Flo0807 I changed a few internal function with theoretically could be used from the outside. The chance is relatively small... should we add it to the upgrade guide?

lib/backpex/fields/has_many.ex
```diff
@@ -341,8 +341,7 @@ defmodule Backpex.Fields.HasMany do
     } = assigns
 
     link =
-      if link_assocs and Map.has_key?(field_options, :live_resource) and
-           LiveResource.can?(assigns, :show, item, live_resource) do
+      if link_assocs and Map.has_key?(field_options, :live_resource) and live_resource.can?(assigns, :show, item) do
```
Flo0807, line 344:
```suggestion
      if link_assocs and live_resource.can?(assigns, :show, item) do
```

The `live_resource` key is required so we can remove this condition. 

I guess, we also have to change this call: `live_resource.can?(assigns, :show, item)` and use the live resource of the association?

```suggestion
      if link_assocs and field_options.live_resource.can?(assigns, :show, item) do
```

(untested)

lib/backpex/live_resource.ex
```diff
@@ -1585,35 +1590,35 @@ defmodule Backpex.LiveResource do
...
-    |> Map.drop([Atom.to_string(module.get_empty_filter_key())])
+    |> Map.drop([Atom.to_string(get_empty_filter_key())])
   end
 
   @doc """
   Returns list of active filters.
   """
-  def get_active_filters(module, assigns) do
-    filters = module.filters(assigns)
+  def active_filters(assigns) do
```
Flo0807, line 1613:
```suggestion
  def active_filters(live_resource, assigns) do
```

Even if it's already included in the assigns, adding the `live_resource` as a param makes it more explicit.

I am unsure if I like it or not...

lib/backpex/live_resource.ex
```diff
@@ -1372,6 +1385,8 @@ defmodule Backpex.LiveResource do
 
   def filter_options(_no_filters_present, _filter_configs), do: %{}
 
+  def get_empty_filter_key, do: @empty_filter_key
```
Flo0807, line 1388:
```suggestion
  def get_empty_filter_key, do: :empty_filter
```

What's the benefit of using a module attribute here? I recently learned that module attributes should not be used for constants. Do we need this key at compile time?

See also `permitted_order_directions`

https://hexdocs.pm/elixir/module-attributes.html#as-compile-time-constants

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+    case Map.get(saved_fields, Atom.to_string(name)) do
+      "true" ->
+        true
+
+      "false" ->
+        false
+
+      _other ->
+        true
+    end
```
Flo0807, line 525:
```suggestion
    case Map.get(saved_fields, Atom.to_string(name)) do
      "true" -> true
      "false" -> false
      _other -> true
    end
```

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+  end
+
+  def assign_items(socket) do
+    %{
+      live_resource: live_resource,
+      fields: fields
+    } = socket.assigns
```
Flo0807, line 532:
```suggestion
    %{live_resource: live_resource, fields: fields} = socket.assigns
```

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+            {key, metric}
+        end
+      end)
+
+    socket
+    |> assign(metrics: metrics)
+  end
+
+  @impl Phoenix.LiveView
+  def render(%{live_action: action} = assigns) when action in [:show, :show_edit] do
```
Flo0807, line 589:
```suggestion
  def render(%{live_action: action} = assigns) when action in [:show] do
```

What is `show_edit` even for? It is the only occurrence.

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+    |> assign(:page_title, socket.assigns.plural_name)
+    |> apply_index()
+    |> assign(:item, nil)
+  end
+
+  defp apply_action(socket, :edit) do
+    %{
+      live_resource: live_resource,
+      singular_name: singular_name
+    } = socket.assigns
```
Flo0807, line 625:
```suggestion
    %{live_resource: live_resource, singular_name: singular_name} = socket.assigns
```

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+    |> assign(:page_title, Backpex.translate({"Edit %{resource}", %{resource: singular_name}}))
+    |> assign(:item, item)
+    |> assign_changeset(fields)
+  end
+
+  defp apply_action(socket, :show) do
+    %{
+      live_resource: live_resource,
+      singular_name: singular_name
+    } = socket.assigns
```
Flo0807, line 645:
```suggestion
    %{live_resource: live_resource, singular_name: singular_name} = socket.assigns
```

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+    |> apply_show_return_to(item)
+  end
+
+  defp apply_action(socket, :new) do
+    %{
+      live_resource: live_resource,
+      schema: schema,
+      create_button_label: create_button_label
+    } = socket.assigns
+
```
Flo0807, line 666:
```suggestion
    %{live_resource: live_resource, schema: schema, create_button_label: create_button_label} = socket.assigns

```

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+    %{
+      live_resource: live_resource,
+      singular_name: singular_name
+    } = socket.assigns
+
+    fields = live_resource.fields |> filtered_fields_by_action(socket.assigns, :edit)
+    primary_value = URI.decode(socket.assigns.params["backpex_id"])
+    item = Resource.get!(primary_value, socket.assigns, live_resource)
+
+    unless live_resource.can?(socket.assigns, :edit, item), do: raise(Backpex.ForbiddenError)
```
Flo0807, line 631:
```suggestion
    if not live_resource.can?(socket.assigns, :edit, item), do: raise(Backpex.ForbiddenError)
```

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+    %{
+      live_resource: live_resource,
+      singular_name: singular_name
+    } = socket.assigns
+
+    fields = live_resource.fields() |> filtered_fields_by_action(socket.assigns, :show)
+    primary_value = URI.decode(socket.assigns.params["backpex_id"])
+    item = Resource.get!(primary_value, socket.assigns, live_resource)
+
+    unless live_resource.can?(socket.assigns, :show, item), do: raise(Backpex.ForbiddenError)
```
Flo0807, line 651:
```suggestion
    if not live_resource.can?(socket.assigns, :show, item), do: raise(Backpex.ForbiddenError)
```

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+  end
+
+  defp apply_action(socket, :new) do
+    %{
+      live_resource: live_resource,
+      schema: schema,
+      create_button_label: create_button_label
+    } = socket.assigns
+
+    unless live_resource.can?(socket.assigns, :new, nil), do: raise(Backpex.ForbiddenError)
```
Flo0807, line 667:
```suggestion
    if not live_resource.can?(socket.assigns, :new, nil), do: raise(Backpex.ForbiddenError)
```

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+    %{live_resource: live_resource} = socket.assigns
+
+    id =
+      socket.assigns.params["backpex_id"]
+      |> URI.decode()
+      |> String.to_existing_atom()
+
+    action = live_resource.resource_actions()[id]
+
+    unless live_resource.can?(socket.assigns, id, nil), do: raise(Backpex.ForbiddenError)
```
Flo0807, line 690:
```suggestion
    if not live_resource.can?(socket.assigns, id, nil), do: raise(Backpex.ForbiddenError)
```

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+  end
+
+  defp apply_index(socket) do
+    %{
+      live_resource: live_resource,
+      schema: schema,
+      params: params
+    } = socket.assigns
+
+    unless live_resource.can?(socket.assigns, :index, nil), do: raise(Backpex.ForbiddenError)
```
Flo0807, line 733:
```suggestion
    if not live_resource.can?(socket.assigns, :index, nil), do: raise(Backpex.ForbiddenError)
```

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+    |> apply_index_return_to()
+  end
+
+  defp assign_changeset(socket, fields) do
+    %{
+      item: item,
+      changeset_function: changeset_function,
+      live_action: live_action
+    } = socket.assigns
+
```
Flo0807, line 796:
```suggestion
  defp assign_changeset(socket, fields) do
    %{item: item, changeset_function: changeset_function, live_action: live_action} = socket.assigns

```

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+  defp default_attrs(_live_action, _fields, _assigns), do: %{}
+
+  defp maybe_redirect_to_default_filters(%{assigns: %{filters_changed: false}} = socket) do
+    %{
+      live_resource: live_resource,
+      query_options: query_options,
+      params: params,
+      filters: filters
+    } = socket.assigns
```
Flo0807, line 838:
```suggestion
    %{live_resource: live_resource, query_options: query_options, params: params, filters: filters} = socket.assigns
```

lib/backpex/live_resource.ex
```diff
@@ -1231,6 +447,803 @@ defmodule Backpex.LiveResource do
...
+    |> assign(:query_options, new_query_options)
+    |> assign_items()
+    |> maybe_assign_metrics()
+  end
+
+  defp update_item(socket, item) do
+    %{
+      live_resource: live_resource,
+      live_action: live_action
+    } = socket.assigns
```
Flo0807, line 1174:
```suggestion
    %{live_resource: live_resource, live_action: live_action} = socket.assigns
```

Flo0807 (changes requested):

lib/backpex/live_resource.ex
pehbehbeh: I think its in line with the rest of the application if we just use the one from `assigns`.

lib/backpex/live_resource.ex
```diff
@@ -248,806 +245,28 @@ defmodule Backpex.LiveResource do
...
 
       @impl Phoenix.LiveView
-      def handle_event("close-modal", _params, socket) do
-        socket =
-          socket
-          |> push_patch(to: socket.assigns.return_to)
-
-        {:noreply, socket}
-      end
+      def handle_event(event, params, socket), do: LiveResource.handle_event(event, params, socket)
```
Flo0807, line 263: @pehbehbeh this will catch all events, and users are no longer able to define own `handle_event` functions in their LiveResources. Same with `handle_info` etc.

PR #665, opened 2024-11-12
Add ‘rows' attribute to the textarea, with a default value of 2 rows.

guoxinghuang:
Add ‘rows' attribute to the textarea, with a default value of 3 rows.

### DIFF
--- a/lib/backpex/fields/textarea.ex
+++ b/lib/backpex/fields/textarea.ex
@@ -34,7 +34,7 @@ defmodule Backpex.Fields.Textarea do
         <BackpexForm.input
           type="textarea"
           field={@form[@name]}
-          rows={@field_options[:rows] || 3}
+          rows={@field_options[:rows] || 2}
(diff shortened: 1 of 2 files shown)


lib/backpex/fields/textarea.ex
```diff
@@ -34,6 +34,7 @@ defmodule Backpex.Fields.Textarea do
         <BackpexForm.input
           type="textarea"
           field={@form[@name]}
+          rows={@field_options[:rows] || 3}
```
Flo0807, line 37:
```suggestion
          rows={@field_options[:rows] || 2}
```

According to the mdn web docs, the row attribute defaults to 2. so let's use 2 for the default, too.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#rows

guoxinghuang: Perfect 🤣

Flo0807:
Can you add `row` to the list of options (on top of the `Backpex.Fields.Textarea` module) with a brief description? @guoxinghuang

PR #666, opened 2024-11-14
Support ash destroy action

Flo0807:


lib/backpex/adapters/ash.ex
```diff
@@ -75,8 +75,17 @@ if Code.ensure_loaded?(Ash) do
...
+      config = live_resource.config(:adapter_config)
+      primary_key = live_resource.config(:primary_key)
+
+      ids = Enum.map(items, &Map.fetch!(&1, primary_key))
+
+      config[:resource]
+      |> Ash.Query.filter(^Ash.Expr.ref(primary_key) in ^ids)
+      |> Ash.bulk_destroy!(:destroy, %{})
+
+      :ok
```
Flo0807, line 88: How do we want to catch errors? Currently, the item action, e.g., Delete, uses `try rescue` and pattern matches on the error. I don't like this, as the item action itself should not figure out why the deletion failed. In the case of Ash, the bulk delete might contain errors (in the result struct), even if it did not raise an error. I suggest moving the error handling to the adapter.  @pehbehbeh

lib/backpex/item_actions/item_action.ex
```diff
@@ -22,6 +22,8 @@ defmodule Backpex.ItemAction do
   @callback fields() :: list()
 
   @doc """
+  TODO: rename to `init_schema`
```
Flo0807, line 25: I want to rename `init_change`. You might think that this function passes the result as the change / params / attrs into the changeset, but it is used as the source changeset data. WDYT, is `init_schema` a good name? @pehbehbeh

lib/backpex/live_components/form_component.ex
```diff
@@ -81,9 +83,12 @@ defmodule Backpex.FormComponent do
...
       |> put_upload_change(socket, :validate)
 
+    metadata = Resource.build_changeset_metadata(socket.assigns, target)
+
     changeset =
-      item_action_types
-      |> Resource.change(change, fields, assigns, live_resource, target: target)
+      init_schema
+      |> changeset_function.(change, metadata)
+      |> Map.put(:action, :validate)
```
Flo0807, line 91: Still not sure where to place the code that calls the changeset function and if we should move it to another module 🤔 I also did not include special calls like `before_changeset`. We even state that we do not support advanced fields like `HasMany` in item actions in our docs (of course, we should change this in the future).

lib/backpex/item_actions/item_action.ex
Flo0807: See https://github.com/naymspace/backpex/pull/672

lib/backpex/adapters/ash.ex
Flo0807: We decided to use `{:ok, deleted_items}` for now. We plan to improve error handling in the future so that the item action can match on `:ok` or `:error` without the need to use `try ... rescue`

lib/backpex/item_actions/item_action.ex
```diff
@@ -111,10 +113,28 @@ defmodule Backpex.ItemAction do
...
+  end
+
+  @doc """
+  Checks whether item action has form.
+  """
+  def has_form?(item_action) do
+    module = Map.fetch!(item_action, :module)
+
+    module.fields() != []
+  end
```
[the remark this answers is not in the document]
pehbehbeh: Lets keep it that way for now.

pehbehbeh (approved):

PR #667, opened 2024-11-14
standardize and improve resource and adapter modules

pehbehbeh:
- standardize functions in `Backpex.Resource` (parameters and result tuples)
- standardize functions in `Backpex.Adapter` (parameters incl. typespecs)
- implement bang function (only `get!/1` for now) in `Backpex.Resource` instead of specific adapters
- simplify stuff by reducing parameters


lib/backpex/exceptions.ex
```diff
@@ -1,16 +1,9 @@
-defmodule Backpex.NoResourceError do
```
pehbehbeh, line 1: `NoResourceError` is not used anymore, as far as I can see.

lib/backpex/resource.ex
```diff
@@ -40,35 +40,37 @@ defmodule Backpex.Resource do
...
-  def get(primary_value, assigns, live_resource) do
-    adapter = live_resource.config(:adapter)
-
-    adapter.get!(primary_value, assigns, live_resource)
+  def get!(primary_value, assigns, live_resource) do
+    case get(primary_value, assigns, live_resource) do
+      {:ok, nil} -> raise Backpex.NoResultsError
+      {:ok, result} -> result
+      {:error, _error} -> raise Backpex.NoResultsError
+    end
```
pehbehbeh, line 73: @Flo0807 Like this? I think we should work with result tuples everywhere (like Ash for example).

Flo0807: 👍

demo/lib/demo_web/live/ticket_live.ex
```diff
@@ -24,8 +24,8 @@ defmodule DemoWeb.TicketLive do
       <Backpex.HTML.CoreComponents.icon name="hero-information-circle" class="h-5 w-5" />
       <p>
         This resource uses the <strong>Ash adapter</strong>, which is currently in a very early alpha stage.
-        Currently, only <strong>index</strong> and <strong>show</strong> are functional in a very basic form.
-        We are working on supporting more Backpex features in the future.
+        Currently, only <strong>index</strong>, <strong>show</strong> and <strong>delete</strong> are functional in a
+        very basic form. We are working on supporting more Backpex features in the future.
```
Flo0807, line 28: I'd keep a hint that this resources uses Ash. You can't tell just from looking at the page in the demo and might wonder why this resource only supports index, show and edit.

Flo0807 (approved):

lib/backpex/live_resource.ex
```diff
@@ -328,9 +319,6 @@ defmodule Backpex.LiveResource do
       @impl Backpex.LiveResource
       def search_placeholder, do: Backpex.translate("Search")
 
-      @impl Backpex.LiveResource
-      def item_query(query, _live_action, _assigns), do: query
```
Flo0807, line 332: This will no longer add fallback (cover all) `item_query/3` functions to your LiveResources. I added a hint to the v0.11 upgrade guide that you have to cover all cases with `item_query/3`. We had this error in our demo, too. The User edit, show and new view crashed because of this.

PR #672, opened 2024-11-26
Rename item and resource action `init_change/1` callback to `base_schema/1`

Flo0807:
The term `init_change` is confusing because the result is used as the base schema / item for the changeset function.

### DIFF
--- a/lib/backpex/live_resource.ex
+++ b/lib/backpex/live_resource.ex
@@ -672,7 +672,7 @@ defmodule Backpex.LiveResource do
     if not live_resource.can?(socket.assigns, id, nil), do: raise(Backpex.ForbiddenError)
 
     changeset_function = &action.module.changeset/3
-    item = action.module.base_item(socket.assigns)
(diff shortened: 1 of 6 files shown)


lib/backpex/live_resource.ex
```diff
@@ -1173,19 +1177,17 @@ defmodule Backpex.LiveResource do
...
-    changeset =
-      init_change
-      |> Ecto.Changeset.change()
-      |> changeset_function.(%{}, metadata)
+    changeset = changeset_function.(base_schema, %{}, metadata)
 
     socket =
       socket
-      |> assign(:item_action_types, init_change)
+      |> assign(:item, base_schema)
```
Flo0807, line 1190: Internally, the `form_component` treats the `base_schema` as the `item` (we use the same `form_component` like in the `edit` and `create` view). That's why I kept the name `item`. From the outside it has nothing to do with the item, which is why I chose the name `base_schema`.

PR #673, opened 2024-11-26
Add `init_params/1` callback for item actions

Flo0807:


lib/backpex/live_resource.ex
```diff
@@ -1174,12 +1174,13 @@ defmodule Backpex.LiveResource do
...
   defp open_action_confirm_modal(socket, action, key) do
     base_item = action.module.base_item(socket.assigns)
+    init_params = action.module.init_params(socket.assigns)
 
     changeset_function = &action.module.changeset/3
 
     metadata = Resource.build_changeset_metadata(socket.assigns)
 
-    changeset = changeset_function.(base_item, %{}, metadata)
+    changeset = changeset_function.(base_item, init_params, metadata)
```
Flo0807, line 1183: @pehbehbeh We should combine this with the `default` values users can define for fields. Currently, we do not use these values in Item Actions (we do in Resource Actions: see https://github.com/naymspace/backpex/blob/develop/lib/backpex/live_resource.ex#L794)

What are your thoughts on using the `init_params` if defined and otherwise the `default` values (same for resource actions)? 

Something like this:
```elixir
init_params = action.module.init_params(socket.assigns) || build_default_params
```

If both `init_params` and "default" field values are not present an empty map is being used.

pehbehbeh: Sounds reasonable!

pehbehbeh (approved):

PR #675, opened 2024-11-27
Change expected return values of item and resource action `handle` functions

Flo0807:


lib/backpex/item_actions/item_action.ex
```diff
@@ -75,9 +75,19 @@ defmodule Backpex.ItemAction do
...
+
+  If `{:error, changeset}` is returned, the changeset is used to update the form to display the errors. Note that Backpex already validates the form for you.
+  Therefore it is only necessary in rare cases to perform additional validation and return a changeset from `c:handle/3`.
+  For example, if you are building a duplicate action and can only check for a unique constraint when inserting the duplicate element.
+
+  You are only allowed to return `{:error, changeset}` if the action has a form. Otherwise Backpex will throw an ArgumentError.
   """
   @callback handle(socket :: Phoenix.LiveView.Socket.t(), items :: list(map()), params :: map() | struct()) ::
-              {:noreply, Phoenix.LiveView.Socket.t()} | {:reply, map(), Phoenix.LiveView.Socket.t()}
+              {:noreply, Phoenix.LiveView.Socket.t()} | {:error, Ecto.Changeset.t()}
```
Flo0807, line 90: Thought about turning `{:noreply, socket}` into `{:ok, socket}` because the message type is not used anymore 🤔 What do you think? @pehbehbeh

pehbehbeh: Sounds good.

pehbehbeh (approved):

PR #696, opened 2024-12-11
Streamline live resource generation by inspecting the resource Ecto schema

tmbb:
Introspect the ecto schema and automatically inflect the schema name to infer the singular name, plural name and the pubsub event name. The following exmaple is copied from the `@moduledoc`. This makes it very succint to use Backpex to add an admin layer to a number of Ecto schemas with very little code.

To use this functionality in your app, first define a new module in your application which encapsulates data that is common to all resources (repo, pubsub and layouts)

    defmodule MyAppWeb.LiveResource do
      use Backpex.AutoLiveResource,
        repo: MyApp.Repo,
        layout: {MyAppWeb.Layouts, :admin},
        pubsub: MyApp.PubSub
    end

Now, you can define a new live resource using the module above

    defmodule MyAppWeb.MyResourceLive do
      use MyAppWeb.LiveResource,
        resource: MyApp.MyContext.MyResource
    end

There are currently no integration tests for this (but then again, there are no integration tests for the rest of Backpex).


Flo0807:
Hey! Thanks for the effort, but as I already posted [in the forum thread](https://elixirforum.com/t/backpex-a-highly-customizable-admin-panel-for-phoenix-liveview-applications/64314/41), an `AutoLiveResource` module, which is another abstraction layer for live resources, is not something we want to add to the core of Backpex. We'd like to simplify the setup in the future, but we'd like to do that with dedicated generators.

PR #726, opened 2024-12-23
docs: update backpex custom field example

lewisf:
I was trying to create a custom field and was running into all sorts of errors because other necessary dependencies were not available.

This seems to get it closer to working as expected


lewisf:
FWIW even with this I'm still getting errors on the minimal custom field example like

`Backpex.HTML.Form.input/1 is undefined or private`
`Backpex.Field.translate_error_fun/2 is undefined or private`

PR #761, opened 2025-01-10
Fix empty upload not showing error on save

thomasfortes:
This fix the upload field not showing an error when saving by removing the `_unused_field` from the changes and letting Phoenix handle the error.

It still does not solve the caveat that removing an image does not show an error until you interact with another field (but sometimes it does when the user already has an image and you remove it).


Flo0807:
🎉🚀

PR #767, opened 2025-01-14
Add demo label to Dockerfile updates

Flo0807:


Flo0807:
@krns This adds the `demo` label to Dockerfile and `compose.yml` updates because the Dockerfile / `compose.yml` is mainly used for the Demo / dev project and not relevant for the Backpex package itself.

I also thought about automerging patch updates (Elixir packages) for the demo project. What do you think? Would this save some maintenance work for you?

krns:
> I also thought about automerging patch updates (Elixir packages) for the demo project. What do you think? Would this save some maintenance work for you?

Let's try it out. What do you think about adding the auto-merge setting in #538?

PR #783, opened 2025-01-17
introduce ok/1 and noreply/1 helpers

pehbehbeh:


lib/backpex/live_resource.ex
```diff
@@ -1223,8 +1204,7 @@ defmodule Backpex.LiveResource do
         |> assign(action_to_confirm: nil)
         |> assign(selected_items: [])
         |> assign(select_all: false)
-
-        {:noreply, socket}
+        |> noreply()
```
Flo0807, line 1207: We did not reassigned the socket before 🤔

Flo0807 (approved):

lib/backpex/live_resource.ex
pehbehbeh: Yeah... I was wondering about that one, too. I checked the item actions and everything still works. And it makes no sense to assign something there and not "return" it, right? 🤔  Otherwise the assigns would have no effect and could be removed?

Flo0807: > And it makes no sense to assign something there and not "return" it, right? 🤔 Otherwise the assigns would have no effect and could be removed?

Yes, I wonder why item actions fully work without resetting these values.

PR #794, opened 2025-01-24
Add external upload support for the Upload field.

thomasfortes:
I've been using a custom upload field with support for external uploads for a while, and after upgrading `Backpex`, I refactored my field and decided to extract the minimal functionality needed to support the existing Upload field.

The field works out of the box with Cloudflare R2 using the functions provided in the Phoenix LiveView documentation under [Direct to S3-Compatible External Uploads](https://hexdocs.pm/phoenix_live_view/external-uploads.html#direct-to-s3-compatible), but I would be more confident with the code if someone that uses S3 (or other external upload platforms) could test it as well.

If this is something that you're interested I can also update the documentation in this PR.

### DIFF
--- a/lib/backpex/fields/upload.ex
+++ b/lib/backpex/fields/upload.ex
@@ -150,7 +150,7 @@ defmodule Backpex.Fields.Upload do
 
       **Examples**
 
-      This function have the same format as the ones available at the Phoenix Liveview
+      This function has the same format as those available in the Phoenix LiveView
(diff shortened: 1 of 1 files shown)


Flo0807:
Hey @thomasfortes, thanks for the contribution! 🚀

> If this is something that you're interested I can also update the documentation in this PR.

Yes, we are definitely interested in this, so it would be nice to have a minimal documentation for a direct S3 upload in our docs.

PR #818, opened 2025-01-30
Replace Alpine.js with JS hooks

pehbehbeh:
This completely removes Alpine.js dependency.

- Trigger from change (validation) on upload entry cancel via JS hook and server sent event. (Current Alpine.js solution was broken btw.)
- Trigger drag hover styles on upload dropzones via JS hook.
- Use DaisyUI `drawer` component for responsive sidebar.
- Use DaisyUI `collapse` component for sidebar sections and use localstorage via custom events.
- Use `sticky` columns with `IntersectionObserver` to dynamically display the column shadow.
- Moves the theme selector stuff to backpex core to ease the installation.


lib/backpex/html/resource/resource_index_table.html.heex
```diff
@@ -94,37 +55,24 @@
...
             :for={{key, action} <- row_item_actions(@item_actions)}
             :if={@live_resource.can?(assigns, key, item)}
-            class="tooltip hover:z-30"
-            data-tip={action.module.label(assigns, item)}
+            type="button"
+            phx-click="item-action"
+            phx-value-action-key={key}
+            phx-value-item-id={Map.get(item, @live_resource.config(:primary_key))}
+            aria-label={action.module.label(assigns, item)}
+            title={action.module.label(assigns, item)}
```
pehbehbeh, line 72: @Flo0807 @krns 

I had to remove the item action tooltips because they cannot break out of the `sticky` container without breaking everything. :-(  Can we deal with that? I would be fine with just a `title` attribute for now.

I guess we could find a solution with the Popover API, but that requires an ID for every single tooltip as far as I can see... maybe something for later?

Everything else works pretty good... I got this working with a `position: sticky;` column and a `IntersectionObserver` to dynamically display the column shadow. 

Bildschirmfoto 2025-02-06 um 00 29 08

assets/js/hooks/_sticky_actions.js
```diff
@@ -0,0 +1,20 @@
...
+ * Hook to determine if a table row is horizontally scrollable (=> actions are sticky).
+ *
+ * Adds a 'stuck' attribute to the sticky element which can then be used for styling.
+ */
+export default {
+  mounted() {
+    const sticky = this.el.querySelector('.sticky')
+
+    this.observer = new IntersectionObserver(
+      ([entry]) => { sticky.toggleAttribute('stuck', entry.intersectionRatio < 1)},
```
pehbehbeh, line 11: I just found another bug... the `stuck` attribute is removed on re-renders (select item or search for example). I'm probably going crazy here... 🥴

assets/js/hooks/_drag_hover.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * This hook adds hover styles on the upload dropzone when dragging a file over it.
+ */
+export default {
+  mounted() {
+    this.dragging = 0
+    this.el.addEventListener("dragenter", () => { this.dragChange(this.dragging + 1) })
+    this.el.addEventListener("dragleave", () => { this.dragChange(this.dragging - 1) })
+    this.el.addEventListener("drop", () => { this.dragChange(0) })
+  },
```
krns, line 10: remove the event listeners on destroy

assets/js/hooks/_theme_selector.js
```diff
@@ -0,0 +1,50 @@
...
+        await fetch(cookiePath, {
+          body: `select_theme=${selectedTheme.value}`,
+          method: 'POST',
+          headers: {
+            'Content-type': 'application/x-www-form-urlencoded',
+            'x-csrf-token': csrfToken
+          }
+        })
+      }
+    })
```
krns, line 40: Remove event listener on destroy

demo/assets/js/app.js
```diff
@@ -27,84 +21,18 @@ window.addEventListener('phx:page-loading-start', _info => topbar.show(250))
...
-    onNodeAdded (node) {
-      // Mimic autofocus for dynamically inserted elements
-      if (node.nodeType === window.Node.ELEMENT_NODE && node.hasAttribute('autofocus')) {
-        node.focus()
-
-        if (node.setSelectionRange && node.value) {
-          const lastIndex = node.value.length
-          node.setSelectionRange(lastIndex, lastIndex)
-        }
-      }
```
krns, line 101: I can't remember why we added this, but I wanted to add a comment in case this wasn't Alpine's reason (I don't see any Alpine specific code like in `onBeforeElUpdated`).

assets/js/hooks/_sticky_actions.js
```diff
@@ -0,0 +1,29 @@
...
+  updated() {
+    this.toggleStuckAttribute()
+  },
+  destroyed() {
+    this.observer.disconnect()
+  },
+  toggleStuckAttribute() {
+    this.sticky.toggleAttribute('stuck', this.stuck)
+  }
+}
```
krns, line 29: Do you know if this works if the item action column has different number of actions per row? With the current implementation this did not work, so we had to make some workarounds. 

The specific use case: We have a resource where items should only be able to be archived (soft deleted). When an item is archived it cannot be edited. This looked like the following:

- show, edit, archive
- show, unarchive

The unequal number of item actions broke the layout, if I remember correctly.

Out workaround is that we display the delete action for archived items so that the number of item actions is the same again:

- show, edit, archive
- show, unarchive, delete

krns: Just checked out the demo of your feature branch and it seems to work:

Screenshot 2025-02-05 at 20 10 59




@Flo0807 Or was our problem different and based on the fact that the number of items per row was changed by the user (when archiving)? And the initial rendering was not the problem at all?

lib/backpex/html/resource/resource_index_table.html.heex
krns: I think that's acceptable. But it would be good if we could create an issue for it and add it to the changelog (and link the ticket).

assets/js/hooks/_sticky_actions.js
Flo0807: > Or was our problem different and based on the fact that the number of items per row was changed by the user (when archiving)? And the initial rendering was not the problem at all?

I guess, the problem only occurred when the number of actions changed dynamically and initial rendering was not a problem at all.

Nevertheless, I tested the new implementation and it seems to fix the issue 🎉

https://github.com/user-attachments/assets/1ccba725-a026-4ad0-89fe-c12b752096cf

krns: Awesome!

Flo0807:
Awesome! 🙌

Flo0807:
I noticed a small issue:

If you select rows, the action column is not highlighted.

@pehbehbeh

lib/backpex/html/resource/resource_index_table.html.heex
```diff
@@ -1,37 +1,16 @@
...
-    },
-  }
-  "
-  x-init="setActionsWidth()"
-  x-on:resize.window="setActionsWidth()"
->
-  <thead class="bg-base-200/50 text-base-content uppercase">
-    <tr>
+<table class="table">
+  <thead class="bg-base-300 text-base-content uppercase">
```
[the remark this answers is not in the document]
Flo0807: I guess, we need to keep darker colors. Otherwise, the opacity results in a transparent background so you can see some text below the action column.

lib/backpex/html/layout.ex
```diff
@@ -398,23 +375,12 @@ defmodule Backpex.HTML.Layout do
...
+    <div phx-mounted={JS.dispatch("backpex:sidebar-section-mounted")} class="collapse collapse-arrow" data-id={@id}>
+      <input type="checkbox" checked phx-change={JS.dispatch("backpex:sidebar-section-toggled")} />
+      <div class={["collapse-title text-sm font-semibold uppercase", @class]}>
+        {render_slot(@label)}
       </div>
-      <div class="flex-col space-y-1" x-show="open" x-transition x-transition.duration.75ms>
+      <div class="collapse-content flex-col space-y-1 px-0">
         {render_slot(@inner_block)}
       </div>
     </div>
```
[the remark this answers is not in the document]
Flo0807: I refactored the sidebar to use the daisyUI menu component:

image

@krns this should allow us to remove the custom sidebar implementations in our projects.

Flo0807:
I added `standard` as a `devDependency` to the Backpex `package.json`. This allows us to lint the JavaScript files in our CI workflow.

lib/backpex/html/layout.ex
```diff
@@ -28,78 +28,50 @@ defmodule Backpex.HTML.Layout do
...
-
-          <div class="w-14 flex-shrink-0">
-            <!-- Force sidebar to shrink to fit close icon -->
-          </div>
+    <div id="backpex-app-shell" class="drawer" phx-hook="BackpexSidebarSections">
+      <input id="menu-drawer" type="checkbox" class="drawer-toggle" />
+      <div class="drawer-content">
+        <div class="bg-base-200 fixed inset-0 -z-10 h-full w-full"></div>
+        <div class={[
+          "menu hidden overflow-y-scroll px-2 pt-5 pb-4 md:fixed md:inset-y-0 md:mt-16 md:block md:w-64",
```
Flo0807, line 36: We set `md:w-64` for the sidebar with on the desktop. For the drawer, we use `w-64`. If I want to override this width, I have to use `md:w-72 w-72`. Took me some while to figure this out. 

This isn't a good dev experience, and we might need some sort of Tailwind Merge algorithm / dependency to make it easy to override certain default Backpex classes.


Simply setting `w-72` won't work here because it's overridden by the `md:w-64` class.

cc @krns @pehbehbeh

PR #832, opened 2025-02-05
Configure additionalBranchPrefix renovate option

Flo0807:


Flo0807:
This prevents Renovate from updating the same package both in Backpex and the demo. This allows us to labels these PRs accordingly.

PR #886, opened 2025-02-20
remove Backpex.Adapter.update/2

pehbehbeh:
This remove the `Backpex.Adapter.update/2` callback for adapters. Updates on single items now use the `update_all/3` callback with a list of the given single item.


lib/backpex/resource.ex
```diff
@@ -119,14 +119,17 @@ defmodule Backpex.Resource do
...
-    item
-    |> change(attrs, fields, assigns, live_resource, Keyword.put(opts, :action, :update))
-    |> adapter.update(live_resource)
-    |> after_save(after_save_fun)
-    |> broadcast("updated", pubsub)
+    updates =
+      item
+      |> change(attrs, fields, assigns, live_resource, Keyword.put(opts, :action, :update))
+      |> Map.get(:changes)
+      |> Map.put(:updated_at, DateTime.utc_now())
```
pehbehbeh, line 126: I am not yet completely satisfied with this change. It was done automatically before. This causes two problems:

- Is there even an `updated_at` on the resource?
- Is `DateTime.utc_now()` the right choice?

Flo0807: We could use Ectos reflection functions to get information about it.

`__schema__(:autogenerate_fields)` might return `[:inserted_at, :updated_at]` and we can also check the type with `__schema__(:type, :updated_at)`.

We can then use pattern matching to add the correct timestamp type.

pehbehbeh: Yeah, was also thinking about that but then this should be part of the Ecto adapter... 🤔

Flo0807: > but then this should be part of the Ecto adapter

yes, sure

pehbehbeh:
@Flo0807 I think that must remain separate. It is not possible to maintain assocs via `update_all´ / `insert_all`:

> When a schema module is given, the entries given will be properly dumped before being sent to the database. If the schema primary key has type :id or :binary_id, it will be handled either at the adapter or the storage layer. However any other primary key type or autogenerated value, like [Ecto.UUID](https://hexdocs.pm/ecto/Ecto.UUID.html) and timestamps, won't be autogenerated when using [insert_all/3](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert_all/3). You must set those fields explicitly. This is by design as this function aims to be a more direct way to insert data into the database without the conveniences of [insert/2](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert/2). This is also consistent with [update_all/3](https://hexdocs.pm/ecto/Ecto.Repo.html#c:update_all/3) that does not handle auto generated values as well.
> 
> It is also not possible to use insert_all to insert across multiple tables, therefore associations are not supported.

https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert_all/3

You get the following error when updating assocs:

```
app-1       | ** (Ecto.QueryError) deps/ecto/lib/ecto/repo/queryable.ex:174: field `posts` in `update` is an association in schema Demo.User in query:
app-1       |
app-1       | from u0 in Demo.User,
app-1       |   where: u0.id in ^[1],
app-1       |   update: [
app-1       |   set: [
app-1       |     posts:
app-1       |       ^[
app-1       |         #Ecto.Changeset<action: :update, changes: %{}, errors: [], data: #Demo.Post<>, valid?: true, ...>,
app-1       |         #Ecto.Changeset<action: :update, changes: %{}, errors: [], data: #Demo.Post<>, valid?: true, ...>
app-1       |       ],
app-1       |     updated_at: ^~U[2025-04-03 10:57:26.919025Z]
app-1       |   ]
app-1       | ],
app-1       |   select: u0
app-1       |
app-1       |     (elixir 1.18.3) lib/enum.ex:2546: Enum."-reduce/3-lists^foldl/2-0-"/3
app-1       |     (elixir 1.18.3) lib/enum.ex:1840: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
app-1       |     (elixir 1.18.3) lib/enum.ex:2546: Enum."-reduce/3-lists^foldl/2-0-"/3
app-1       |     (ecto 3.12.5) lib/ecto/repo/queryable.ex:214: Ecto.Repo.Queryable.execute/4
app-1       |     (backpex 0.12.0) lib/backpex/resource.ex:147: Backpex.Resource.update_all/4
app-1       |     (backpex 0.12.0) lib/backpex/resource.ex:127: Backpex.Resource.update/6
```

Flo0807:
> @Flo0807 I think that must remain separate. It is not possible to maintain assocs via `update_all` / `insert_all`:
> 
> > When a schema module is given, the entries given will be properly dumped before being sent to the database. If the schema primary key has type :id or :binary_id, it will be handled either at the adapter or the storage layer. However any other primary key type or autogenerated value, like [Ecto.UUID](https://hexdocs.pm/ecto/Ecto.UUID.html) and timestamps, won't be autogenerated when using [insert_all/3](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert_all/3). You must set those fields explicitly. This is by design as this function aims to be a more direct way to insert data into the database without the conveniences of [insert/2](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert/2). This is also consistent with [update_all/3](https://hexdocs.pm/ecto/Ecto.Repo.html#c:update_all/3) that does not handle auto generated values as well.
> > It is also not possible to use insert_all to insert across multiple tables, therefore associations are not supported.
> 
> https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert_all/3
> 
> You get the following error when updating assocs:
> 
> ```
> app-1       | ** (Ecto.QueryError) deps/ecto/lib/ecto/repo/queryable.ex:174: field `posts` in `update` is an association in schema Demo.User in query:
> app-1       |
> app-1       | from u0 in Demo.User,
> app-1       |   where: u0.id in ^[1],
> app-1       |   update: [
> app-1       |   set: [
> app-1       |     posts:
> app-1       |       ^[
> app-1       |         #Ecto.Changeset<action: :update, changes: %{}, errors: [], data: #Demo.Post<>, valid?: true, ...>,
> app-1       |         #Ecto.Changeset<action: :update, changes: %{}, errors: [], data: #Demo.Post<>, valid?: true, ...>
> app-1       |       ],
> app-1       |     updated_at: ^~U[2025-04-03 10:57:26.919025Z]
> app-1       |   ]
> app-1       | ],
> app-1       |   select: u0
> app-1       |
> app-1       |     (elixir 1.18.3) lib/enum.ex:2546: Enum."-reduce/3-lists^foldl/2-0-"/3
> app-1       |     (elixir 1.18.3) lib/enum.ex:1840: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
> app-1       |     (elixir 1.18.3) lib/enum.ex:2546: Enum."-reduce/3-lists^foldl/2-0-"/3
> app-1       |     (ecto 3.12.5) lib/ecto/repo/queryable.ex:214: Ecto.Repo.Queryable.execute/4
> app-1       |     (backpex 0.12.0) lib/backpex/resource.ex:147: Backpex.Resource.update_all/4
> app-1       |     (backpex 0.12.0) lib/backpex/resource.ex:127: Backpex.Resource.update/6
> ```

alright

PR #887, opened 2025-02-20
make pubsub config optional

### LINKED ISSUE #335: LiveResource: default pubsub settings
All pubsub settings could be optional:

- `pubsub`: use `:pubsub_server` option of the endpoint?
- `topic`: derive from resource name?
- `event_prefix`: derive from resource name?

pehbehbeh:
closes #335


guides/get_started/installation.md
```diff
@@ -6,6 +6,16 @@ The following guide will help you to install Backpex in your Phoenix application
...
 
+## Global configuration
+
+Set the PubSub server of your application in your `config.exs`:
+
+```elixir
+config :backpex, :pubsub_server, MyApp.PubSub
+```
+
+Otherwise you need to configure it for each `LiveResource` individually.
```
Flo0807, line 17: We should add that this will generate PubSub topics and prefixes automatically. Also, a good idea to add a short explanation on how they are generated.

guides/live_resource/listen-to-pubsub-events.md
```diff
@@ -6,7 +6,7 @@ As mentioned in the [installation guide](get_started/installation.md) you need t
 
 You can listen for Backpex PubSub events by implementing the Phoenix.LiveView [`handle_info/2`](Phoenix.LiveView.html#c:handle_info/2) callback in your *LiveResource* module.
 
-We assume you configured PubSub for your Posts *LiveResource* like this:
+You may overwrite the PubSub configuration for your Posts *LiveResource* like this:
```
Flo0807, line 9: We should add the same explanation here, too. See installation guide.

guides/upgrading/v0.11.md
```diff
@@ -49,4 +49,24 @@ use Backpex.LiveResource,
...
+
+```elixir
+    pubsub: [
+        name: Demo.PubSub
+        topic: "posts",
+        event_prefix: "post_"
+    ]
+```
+
+In order to make it work, you need to set your PubSub server in your config:
```
Flo0807, line 66: If we add more information to the "listen to pubsub events" guide, then we could add a link to it here. Something like "You can find more information on how we generate PubSub topics and prefixes automatically here"

lib/backpex/live_resource.ex
```diff
@@ -1215,12 +1217,43 @@ defmodule Backpex.LiveResource do
...
+    |> Kernel.<>("_")
+  end
+
+  defp resource_name(live_resource) do
+    live_resource
+    |> Module.split()
+    |> List.last()
+    |> String.trim_trailing("Live")
+    |> String.downcase()
+  end
```
Flo0807, line 1249:
```suggestion
  defp resource_name(live_resource) do
    live_resource
    |> Module.split()
    |> List.last()
    |> String.downcase()
    |> String.trim_trailing("live")
  end
```

Still, not sure about this. It could lead to some bugs that are difficult to debug. For example, if you don't stick to the `<Resource>Live` naming convention. Then it can happen that multiple LiveResources generate the same PubSub name (imagine `User.IndexLive` and `Post.IndexLive`).

That's also the reason I suggested more explanation in our guides on how we generate PubSub config values automatically.

Flo0807 (changes requested):

lib/backpex/live_resource.ex
Flo0807: What are your thoughts on using `singular_name/0` and `plural_name/0`?

pehbehbeh: As discussed: We are going to use the live resource module name as a default channel.

guides/live_resource/listen-to-pubsub-events.md
```diff
@@ -1,42 +1,52 @@
 # Listen to PubSub Events
 
-As mentioned in the [installation guide](get_started/installation.md) you need to configure PubSub for your *LiveResources*. Backpex will use the configuration to publish `deleted`, `updated` and `created` events. Backpex will listen to these events and update the UI accordingly. Sometimes you may want to listen to these events and perform some custom actions. For example you want to show a toast to all users currently on the resource that a post has been created.
+As mentioned in the [installation guide](get_started/installation.md) are able to configure PubSub events for your each
+*LiveResources* individually. Backpex will use the configuration to publish `deleted`, `updated` and `created` events.
```
Flo0807, line 4:
```suggestion
As mentioned in the [installation guide](get_started/installation.md) you are able to configure PubSub events for each
*LiveResources* individually. Backpex will use the configuration to publish `deleted`, `updated` and `created` events.
```

guides/live_resource/listen-to-pubsub-events.md
```diff
@@ -1,42 +1,52 @@
...
-You can listen for Backpex PubSub events by implementing the Phoenix.LiveView [`handle_info/2`](Phoenix.LiveView.html#c:handle_info/2) callback in your *LiveResource* module.
-
-We assume you configured PubSub for your Posts *LiveResource* like this:
+You may overwrite the PubSub configuration for your Posts *LiveResource* like this:
 
 ```elixir
 use Backpex.LiveResource,
     ...,
     pubsub: [
         name: Demo.PubSub
```
Flo0807, line 17:
```suggestion
        server: Demo.PubSub
```

Flo0807 (approved):

PR #951, opened 2025-03-15
improve modal component

### LINKED ISSUE #317: Use dialog element for all modals
We probably should use the HTML `dialog` element for all modals.

See also daisyUI implementation: https://daisyui.com/components/modal/

This would also solve a current problem: The backdrop is currently scrollable while a modal is open.

pehbehbeh:
Improvements to modal component:

- add `:id` attribute to use multiple modals on the same page
- use DaisyUI modal classes
- add `:class` and `:box_class` attributes to modify the layout
- introduce `:on_cancel` attribute for flexible cancel events (instead of `:close_event_name` and `:target`)
- use `.focus_wrap` component for improved accessibility

Added benefits:

- no server roundtrip when closing resource action modals (removed `close-modal` event)
- no server roundtrip when opening/closing HasManyThrough modal

fixes #317


lib/backpex/fields/has_many_through.ex
```diff
@@ -401,38 +396,6 @@ defmodule Backpex.Fields.HasManyThrough do
...
-  @impl Phoenix.LiveComponent
-  def handle_event("cancel-relational", _params, socket) do
-    %{association: association, return_to_change: return_to_change} = socket.assigns
-
-    put_assoc(association.pivot.field, return_to_change)
-
-    socket
-    |> assign(edit_relational: nil)
-    |> noreply()
-  end
```
pehbehbeh, line 434: I removed the option to cancel an item to simplify a lot here... I suppose that's open for discussion. Modal close, click-away and apply are all the same basically.

WDYT? @Flo0807 @krns

Flo0807: I guess, it is okay. The best case would be to add the association only when the user clicks "Apply", but that makes it really complex.

Flo0807:
Does this PR also fixes:

> The backdrop is currently scrollable while a modal is open.

- https://github.com/naymspace/backpex/issues/317

@pehbehbeh

pehbehbeh:
> Does this PR also fixes?
> 
> > The backdrop is currently scrollable while a modal is open.
> 
> * [Use dialog element for all modals #317](https://github.com/naymspace/backpex/issues/317)
> 
> @pehbehbeh

Yes

PR #954, opened 2025-03-17
Fix some Backpex PubSub events are not handled in certain scenarios

### LINKED ISSUE #953: "Save and Continue Editing" button causes runtime error
Sorry for not using the given template but this should be a simple issue.

"Save and Continue Editing" triggers a consistent bug because the `socket.assigns.live_action` is `:edit` and this case is not handled.

Literally just click this button and see the runtime error.

Using v 0.11.0

peaceful-james:
Fixes #953


Flo0807:
Hey,

the `update_item` action has no effect on edit, as the `edit` live action is not handled there.

Therefore, I'd prefer adding a *catch-all* case which just returns the socket.

```elixir
defp handle_backpex_info({_event, _item}, socket) do
  {:noreply, socket}
end
```

This also fixes future errors where the Backpex events are not handled.

peaceful-james:
@Flo0807 that is done now

Flo0807:
> @Flo0807 that is done now

Awesome 🙌

PR #920, opened 2025-03-03
Upgrade to Tailwind 4 and daisyUI 5

IdoLeshkowitz:
This PR adjusts backpex to tw4 and daisyui 5. This includes migrations to the core library components, as well as to the demo.

PR #957, opened 2025-03-17
Add `backpex.install` command

### LINKED ISSUE #980: `mix backpex.install` - installs backpex

phyr97:
closes #980 

This is the starting point for a backpex installer with [igniter](https://github.com/ash-project/igniter). You can use `mix igniter.install backpex` to add backpex to your project and have almost everything set up without any further work. More information about installing with igniter can be found in its [docs](https://hexdocs.pm/igniter/readme.html#installation). 

The installer requires the new tailwind and daisyui update ([#920](https://github.com/naymspace/backpex/pull/920)), as it greatly reduces the pain of the tailwind and daisyui setup (no tailwind.config.js :partying_face:). 

### ToDo
- [x] Install the backpex hooks via [igniterjs](https://github.com/ash-project/igniter_js)
    - requires igniterjs to be installed in the user's project
- [x] Install daisyui using npm command
    - checks if npm is installed and asks the user if the installer is allowed to install daisyui
- [x] add backpex to `.formatter.exs
- [x] configure pubsub server in `config.exs
- [x] add `backpex_routes` to `router.exs`.
- [x] add `admin.html.heex` to layouts folder
- [x] add `@source "../../deps/backpex/**/*.*ex";` to `app.css`
- [x] warn user if `tailwindcss/forms` plugin exists in `app.css`
    - can be removed by the generator if the user confirms it
- [x] warn user if `bg-white` class is used on `body` tag
    - can be removed by the generator if the user confirms it
- [x] write docs
- [x] test manually with new phoenix project
- [x] improve code
- [x] fix credo warnings and possible bugs discovered while testing
- [ ] write tests for the generator with `Igniter.Test`


lib/mix/tasks/backpex.install.ex
```diff
@@ -0,0 +1,309 @@
...
+      if Igniter.exists?(igniter, root_layout_path) do
+        Igniter.update_file(igniter, root_layout_path, &maybe_remove_bg_white/1)
+      else
+        Igniter.add_warning(igniter, "root.html.heex not found at #{root_layout_path}")
+      end
+    end
+
+    defp maybe_remove_bg_white(source) do
+      root_layout_content = Rewrite.Source.get(source, :content)
+      body_tag_with_bg_white = "<body class=\"h-full bg-white\">"
```
Flo0807, line 249: In a new Phoenix project, there is no `h-full` so we should check for `"<body class=\"bg-white\">"`

See https://github.com/phoenixframework/phoenix/blob/v1.7/installer/templates/phx_web/components/layouts/root.html.heex#L14

I also noticed that Phoenix removes the `bg-white` class starting with Phoenix 1.8, but I think the check is still useful for older projects.

lib/mix/tasks/backpex.install.ex
```diff
@@ -0,0 +1,309 @@
...
+      else
+        Rewrite.Source.update(source, :content, app_css_content <> "\n#{line}")
+      end
+    end
+
+    # check for tailwind forms plugin
+
+    defp check_for_tailwind_forms_plugin(igniter) do
+      app_css_path = igniter.args.options[:app_css_path]
+      line = "@plugin \"tailwindcss/forms\";"
```
Flo0807, line 204: We might want to remove the semicolon because some formatters remove it. It would make the check more reliable.

lib/mix/tasks/backpex.install.ex
phyr97: Oh, yes. I just tested everything with the demo project, but the next step would be to test it with a clean phoenix project. Thanks for the hint.

lib/mix/tasks/backpex.install.ex
```diff
@@ -0,0 +1,322 @@
...
+          )
+      end
+    end
+
+    defp install_daisyui_via_npm do
+      cmd_opts = [stderr_to_stdout: true, env: [{"PATH", System.get_env("PATH")}]]
+
+      with true <- install_daisyui?(),
+           {_version, 0} <- System.cmd("npm", ["--version"], cmd_opts),
+           {_output, 0} <- System.cmd("npm", ["i", "-D", @daisyui_version], cmd_opts) do
```
Flo0807, line 146: I think it is more common to place the `package.json` into `/assets`so we should probably call this within the `/assets` directory.

lib/mix/tasks/backpex.install.ex
```diff
@@ -0,0 +1,322 @@
...
+    defp install_daisyui(igniter) do
+      app_css_path = igniter.args.options[:app_css_path]
+
+      with false <- Helpers.npm_package_installed?(@daisyui_version),
+           :ok <- install_daisyui_via_npm(),
+           igniter <- Helpers.add_line_to_file(igniter, app_css_path, "@plugin \"daisyui\"") do
+        Igniter.add_notice(igniter, "Installed daisyUI via npm.")
+      else
+        true ->
+          Mix.shell().info("daisyui is already in package.json with the desired version. Skipping.")
```
Flo0807, line 129: What if it is in the `package.json`, but not part of the `app.css`? 🤔

Flo0807:
I cannot start the demo app:

```
app-1       | * sourceror (Hex package)
app-1       |   the :only option for dependency sourceror
app-1       |
app-1       |   > In mix.exs:
app-1       |     {:sourceror, "~> 1.7", [env: :prod, hex: "sourceror", only: [:dev, :test], repo: "hexpm"]}
app-1       |
app-1       |   does not match the :only option calculated for
app-1       |
app-1       |   > In deps/rewrite/mix.exs:
app-1       |     {:sourceror, "~> 1.0", [env: :prod, hex: "sourceror", repo: "hexpm", optional: false]}
app-1       |
app-1       |   Remove the :only restriction from your dep
app-1       | ** (Mix) Can't continue due to errors on dependencies
```

I removed `sourceror` from the list of dependency in `demo/mix.exs` and it works. I think the error occurs because `sourceror` is a dependency of `igniter` (?)

lib/mix/tasks/backpex.install.ex
```diff
@@ -0,0 +1,322 @@
...
+        {igniter, nil} ->
+          Mix.shell().error("Could not find router")
+          igniter
+
+        {igniter, router} ->
+          with {:ok, {igniter, false}} <- Helpers.exists_in_module?(igniter, router, "import Backpex.Router"),
+               {:ok, {igniter, false}} <- Helpers.exists_in_module?(igniter, router, "backpex_routes()"),
+               {:ok, igniter} <-
+                 Module.find_and_update_module(igniter, router, &add_backpex_router_import(&1, igniter)) do
+            Phoenix.add_scope(igniter, "/", "backpex_routes()", arg2: web_module)
```
Flo0807, line 191: We have to make sure we pipe through the `:browser` pipeline in the scope. Otherwise, an error occurs when calling the route because the session is not being fetched.

```elixir
scope "/", DemoWeb do
  pipe_through :browser # <-- this line is missing

  backpex_routes()
end
```

Flo0807 (changes requested):

Flo0807:
> This installer automates most of the steps in this guide.

As the installer does not automate all the steps from our guide, the docs should make it clear where you have to continue.

My idea would be to sort the installation guide to first cover the steps the installer can do for you, and then list all the other steps. Then we can link to the section where you can continue.

Flo0807:
> mix igniter.install backpex

I prepared the docs in https://github.com/naymspace/backpex/pull/1098

lib/mix/tasks/backpex.install.ex
phyr97: I can check in which directory the package.json exists and then call the command there.

lib/mix/tasks/backpex.install.ex
phyr97: Even if daisyui is already installed, the function Helpers.add_line_to_file(igniter, app_css_path, ‘@plugin “daisyui”’) will still be executed. The function also checks whether the line already exists in app.css. Nothing will be written to app.css only if the installation of DaisyUI fails or the user has not allowed it.

lib/mix/tasks/backpex.install.ex
phyr97: I have just tested the generator with a new phoenix project using phoenix version 1.8.0-rc.3 (https://github.com/phoenixframework/phoenix/commit/827dca48faca4ea61651238b3f49ebd8a729c18e). As daisyui now comes with Phoenix by default, installing daisyui with our installer would no longer be necessary. So should we wait for Phoenix 1.8?

Flo0807: > I can check in which directory the package.json exists and then call the command there.

👍

> As daisyui now comes with Phoenix by default, installing daisyui with our installer would no longer be necessary. So should we wait for Phoenix 1.8?

I guess, the installer should also work if you've installed Phoenix with the 1.7 generator and do not have daisyui pre-installed.

It's easy to check if DaisyUI is installed via NPM, right? But can we also check whether it is installed without NPM, as is done in Phoenix 1.8? @phyr97 

I'd prefer to check if daisyui is installed (NPM or *Phoenix 1.8-way*). If not, install daisyui via NPM.

phyr97: > I guess, the installer should also work if you've installed Phoenix with the 1.7 generator and do not have daisyui pre-installed.

The problem with the generated phoenix 1.7 project is, that it comes with an older tailwind version with a tailwind config which is hardly to manage in the installer.

phyr97: FYI :
I check the current tailwind version defined in the config.exs. If the version is >=4.x.x, the installer continues. Otherwise the user gets a warning and is asked if the installer should continue or abort the installation. The dependencies will be installed before this check, so the user have to remove the dependencies afterwards when the installer gets aborted.

phyr97:
I would recommend to test the installer with this command:

`mix igniter.install backpex@git:https://github.com/naymspace/backpex.git@feature/backpex-installer`

You also need to install igniter before using this command. You can use the [archive](https://hexdocs.pm/igniter/readme.html) or install it in the mix.exs

priv/templates/layouts/admin.html.heex
```diff
@@ -0,0 +1,27 @@
...
+          <.icon name="hero-arrow-right-on-rectangle" class="size-5" />
+        </.link>
+      </li>
+    </Backpex.HTML.Layout.topbar_dropdown>
+  </:topbar>
+  <:sidebar>
+    <!-- Example Sidebar Item -->
+    <%!-- <Backpex.HTML.Layout.sidebar_item current_url={@current_url} navigate={"/admin/users"}>
+      <.icon name="hero-users" class="size-5" /> Users
+    </Backpex.HTML.Layout.sidebar_item> --%>
```
Flo0807: The example sidebar item is not copied to the project.

lib/mix/tasks/backpex.install.ex
```diff
@@ -0,0 +1,358 @@
...
+          source
+        end
+      else
+        source
+      end
+    end
+
+    defp remove_bg_white? do
+      Mix.shell().yes?(
+        "A background color at the body could cause issues with the backpex app_shell. Do you want to remove it? See: https://hexdocs.pm/backpex/installation.html#remove-default-background-color"
```
Flo0807:
```suggestion
        "A background color at the body could cause issues with the backpex app_shell component. Do you want to remove it? See: https://hexdocs.pm/backpex/installation.html#remove-default-background-color"
```

lib/mix/tasks/backpex.install.ex
```diff
@@ -0,0 +1,358 @@
...
+      app_css_path = igniter.args.options[:app_css_path]
+
+      with false <- Igniter.exists?(igniter, "assets/vendor/daisyui.js"),
+           false <- Helpers.npm_package_installed?(@daisyui_version),
+           :ok <- install_daisyui_via_npm(),
+           igniter <- Helpers.add_line_to_file(igniter, app_css_path, "@plugin \"daisyui\"") do
+        Igniter.add_notice(igniter, "Installed daisyUI via npm.")
+      else
+        true ->
+          Mix.shell().info("daisyui is already installed. Skipping.")
```
Flo0807:
```suggestion
          Mix.shell().info("daisyUI is already installed. Skipping.")
```

lib/mix/tasks/backpex.install.ex
```diff
@@ -0,0 +1,358 @@
...
+
+    - Sets up [Global Configuration](installation.html#global-configuration) by configuring the PubSub server
+    - Adds [Backpex Hooks](installation.html#backpex-hooks) to your app.js file
+    - Installs [daisyUI](installation.html#daisyui) via npm (with your permission)
+    - Sets up the [formatter configuration](installation.html#setup-formatter)
+    - Adds [Backpex files to Tailwind content](installation.html#add-files-to-tailwind-content)
+    - Adds routes to your router
+    - Creates a default admin layout
+    - Checks for and offers to remove the [default background color](installation.html#remove-default-background-color)
+    - Checks for and offers to remove the [@tailwindcss/forms plugin](installation.html#remove-tailwindcssforms-plugin)
```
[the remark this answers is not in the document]
phyr97: This would remove the reference. I wouldn't change that ^^

priv/templates/layouts/admin.html.heex
phyr97: I just found this: 
https://hexdocs.pm/eex/EEx.html#module-tags

```elixir
<%!-- Comments: they are discarded from source --%>
```

I removed the '%'. It works now

lib/mix/tasks/backpex.install.ex
```diff
@@ -0,0 +1,358 @@
...
+
+            Phoenix.add_scope(igniter, "/", contents, arg2: web_module)
+          else
+            {:ok, {igniter, true}} ->
+              Mix.shell().info("Backpex routes already added")
+              igniter
+
+            {:error, igniter} ->
+              Mix.raise("Failed to add backpex routes")
+              igniter
```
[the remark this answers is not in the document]
phyr97: Added warnings with code examples instead of removing the `igniter`. I think we shouldn't stop the installer here just because the routes can't be added.

PR #964, opened 2025-03-18
Add option to disable "Save & Continue editing" button

Flo0807:


Flo0807:
@krns @pehbehbeh 

What do you think, should we disable or enable the button by default?

krns:
I am favouring disabling the button by default because I think it is very counter-intuitive and I don't see a common scenario where this button is needed.

pehbehbeh (approved):

PR #965, opened 2025-03-18
Fix "Save & Continue editing" button when primary key is not named `id`

### LINKED ISSUE #960: Save & Continue Editing does not work with non `id` primary key
We fetch the `id` value of the item to construct the URL for the "Save & Continue editing" button (see https://github.com/naymspace/backpex/blob/0.11.0/lib/backpex/live_components/form_component.ex#L403)

The button breaks if we have a primary key that is not named `id`.

Flo0807:
closes #960


lib/backpex/live_components/form_component.ex
```diff
@@ -395,12 +395,14 @@ defmodule Backpex.FormComponent do
     end)
   end
 
-  defp return_to_path("continue", _live_resource, _socket, %{current_url: url}, item, :new) do
+  defp return_to_path("continue", live_resource, _socket, %{current_url: url}, item, :new) do
+    primary_value = Map.get(item, live_resource.config(:primary_key))
```
pehbehbeh, line 399: Maybe make this public?

```suggestion
    primary_value = live_resource.primary_value(item)
```

pehbehbeh (approved):

lib/backpex/live_components/form_component.ex
Flo0807: We had a `primary_value(item, live_resource)` function in `Backpex.HTML.Resource` and a `primary_value(socket, item)` function in `Backpex.LiveResource`.

I removed the function in `Backpex.HTML.Resource` and created a `primary_value(item, live_resource)` function in `Backpex.LiveResource` that us used everywhere.

PR #973, opened 2025-03-19
WIP: Add `text_overrides` callback to LiveResource

Flo0807:


Flo0807:
This MR proposes a way of how translations could be handled in Backpex to make it easy to provide overrides on a LiveResouce or even global (config) level.

Translations are now real keys / identifiers (e.g., `form.save`) instead of plain texts like "Save". 

Backpex provides a list of default translations for all keys. The list also contains bindings and a description for each Identifier. We will use this list to generate documentation that lists all message keys with bindings and descriptions.

This way, users can look up all the translation keys of Backpex and can use them to override certain texts.

Overriding texts works by implementing the `text_overrides/0` callback in a LiveResource.

The callback hast to return a map that contains message identifiers as keys and their override / translation as values.

If you want to override the text of the save label on form views, you could define your `text_override` function like this:

```elixir
@impl Backpex.LiveResource
def text_overrides do
  %{
    "form.save" => "Continue",
  }
end
```

During translation, Backpex will merge the overrides with the default translations. Therefore, the button text will be "Continue" in this LiveResource. Other LiveResources will display the default text which is "Save".

If the translation contains bindings (the docs will list any bindings for each translation), the User can use a 1-arity function that will receive the bindings as a parameter:

```elixir
@impl Backpex.LiveResource
def text_overrides do
  %{
    "index.create_resource" => fn bindings ->
      gettext("Create new %{resource}", bindings)
    end
  }
end
```

Users still have to provide a translator function in the config. So Backpex continues to translate default texts with this function.

In the future, we can wrap all default translations in a `gettext_noop` call to be able to extract translations into a `pot` file.

The `text_overrides` makes it easy to hook into the translation process and customize any Backpex texts on a LiveResource level. You could also define your own global config anywhere in your app and merge this with your overrides on a LiveResource level.

WDYT? @krns @pehbehbeh

PR #1037, opened 2025-04-01
Fix soft deleting users in the demo

### LINKED ISSUE #990: `DemoWeb.ItemActions.SoftDelete` does not work properly
We need to fix some issues related to soft deletions in our demo application.

- Deleted users are still displayed on posts
- Deleted users can still be selected in posts
- Show pages of deleted users can still be accessed

thomasfortes:
Fixes #990

- Stops fetching deleted users so they don’t show up in posts.  
- Automatically nullifies `user_id` in posts when a user is soft deleted.  
- Prevents access to `:edit` and `:show` actions for deleted users.


demo/lib/demo_web/live/user_live.ex
```diff
@@ -24,7 +24,7 @@ defmodule DemoWeb.UserLive do
   @impl Backpex.LiveResource
   def can?(_assigns, _action, _item), do: true
 
-  def item_query(query, live_action, _assigns) when live_action in [:index, :resource_action] do
+  def item_query(query, live_action, _assigns) when live_action in [:index, :show, :edit, :resource_action] do
```
Flo0807, line 27:
```suggestion
  def item_query(query, live_action, _assigns) do
```

I think we can remove the guard because it covers all possible actions (?) On `new` it has no effect anyway.

demo/lib/demo_web/item_actions/soft_delete.ex
```diff
@@ -64,7 +64,16 @@ defmodule DemoWeb.ItemActions.SoftDelete do
...
+
+        {:ok, _count} =
+          Backpex.Resource.update_all(items, updates, "deleted", socket.assigns.live_resource)
+
+        # nullify the user_id in the posts owned by the users
+        _nullified_posts =
+          items
+          |> Enum.map(fn item ->
+            Backpex.Resource.update_all(item.posts, [set: [user_id: nil]], "updated", DemoWeb.PostLive)
+          end)
```
Flo0807, line 76: This ties the `soft_delete` action to users, so we might want to rename it to something like `user_soft_delete`. What do you think?

thomasfortes: For this specific case I think renaming is fine since only the User resource uses it, but maybe in the future a way to pass extra parameters to actions would help to implement more generic or complex actions, something that we can access in the `handle` like:

```elixir
  @impl Backpex.LiveResource
  def item_actions(default_actions) do
    default_actions
    |> Keyword.drop([:delete])
    |> Enum.concat(
      user_soft_delete: %{
        module: DemoWeb.ItemActions.UserSoftDelete,
        params: [nilify: [{:posts, DemoWeb.PostLive}]]
      }
    )
  end
```

PR #1090, opened 2025-04-22
Refactor input component to use legend and add help_text field option

Flo0807:


Flo0807:
I added a legend to each fieldset to improve a11y. See https://github.com/phoenixframework/phoenix/issues/6204

Flo0807:
See

PR #1124, opened 2025-05-05
Test a11y with playwright

Flo0807:
- Installs and configures `phoenix_test_playwright` dependency
- Assert for a11y issues on the Category pages


demo/test/support/a11y_assertions.ex
```diff
@@ -0,0 +1,23 @@
...
+          session.frame_id
+          |> Frame.evaluate("axe.run()")
+          |> A11yAudit.Results.from_json()
+
+        A11yAudit.Assertions.assert_no_violations(results)
+
+        session
+      end
+    end
+  end
```
Flo0807, line 22: If this code is not part of a macro `mix compile --force --warnings-as-errors` fails.

> warning: PhoenixTest.Playwright.Frame.evaluate/2 is undefined (module PhoenixTest.Playwright.Frame is not available or is yet to be defined)

> A11yAudit.Assertions.assert_no_violations/1 is undefined (module A11yAudit.Assertions is not available or is yet to be defined)

demo/test/demo_web/browser/category_browser_test.exs
```diff
@@ -0,0 +1,45 @@
+defmodule DemoWeb.Browser.CategoryBrowserTest do
+  use PhoenixTest.Playwright.Case, async: true
+  use DemoWeb, :verified_routes
+  use DemoWeb.A11yAssertions
+
+  import Demo.EctoFactory
+
+  @moduletag :external
```
Flo0807, line 8: `:browser` has some side effects and is not possible, that is why I named it `:external` 🤔

demo/config/runtime.exs
```diff
@@ -2,9 +2,25 @@ import Config
...
+    url: [
+      scheme: get_env("URL_SCHEME", "http"),
+      host: get_env("HOST", "localhost"),
+      port: get_env("URL_PORT", "4000")
+    ],
+    secret_key_base: fetch_env!("SECRET_KEY_BASE"),
+    live_view: [
+      signing_salt: fetch_env!("LIVE_VIEW_SIGNING_SALT")
+    ]
+end
```
Flo0807, line 23: If we configure the endpoint without checking for `:prod` env then this overwrites the endpoint configuration from all other configs like `test.exs`

demo/.env.test
```diff
@@ -1,6 +1,3 @@
 LOGGER_LEVEL=warning
-PORT=4002
 MIX_ENV=test
-SECRET_KEY_BASE=np5uHqUApBqAqesURR/sUlc0HJadx/4eBidnQ/w1bH/L1EiCdNlw1id9M/TrziRq
-LIVE_VIEW_SIGNING_SALT=eYlXHghV/uji6ZNUhq+fb63bNVcps7CC
```
Flo0807, line 5: As we now have a dedicated endpoint configuration in `text.exs` we can remove these variables.

.github/workflows/ci.yml
```diff
@@ -231,6 +230,20 @@ jobs:
...
+
+      - name: Run external test
+        working-directory: demo
+        env:
+          DB_HOSTNAME: localhost
+          DB_USERNAME: postgres
+          DB_PASSWORD: postgres
+          DB_DATABASE: test
+        run: yarn test:external
+        continue-on-error: true
```
Flo0807, line 245: I guess we can let the CI crash on errors if we've fixed all the A11Y issues.

demo/test/demo_web/browser/category_browser_test.exs
pehbehbeh: Maybe `:playwright`?

demo/test/support/a11y_assertions.ex
pehbehbeh: The dependency is missing outside of `MIX_ENV=test`.

`MIX_ENV=test mix compile --force --warnings-as-errors` will work.

pehbehbeh (approved):

demo/test/support/a11y_assertions.ex
Flo0807: ok, I'll keep the `use` block then. Otherwise, we need to wrap the alias/import into a `if Mix.env() == :test do` condition.

PR #1155, opened 2025-05-16
add default values in ecto adapter config

pehbehbeh:
- introduce `adapter_config/1` helper (which returns the validated config) and use it everywhere
- adds default values for `:update_changeset`, `:create_changeset` and `:item_query` in ecto adapter


lib/backpex/adapters/ecto.ex
```diff
@@ -64,15 +65,21 @@ defmodule Backpex.Adapters.Ecto do
   use Backpex.Adapter, config_schema: @config_schema
   import Ecto.Query
 
+  @doc false
+  def default_changeset(item, attrs, _metadata), do: Ecto.Changeset.cast(item, attrs, [])
```
Flo0807, line 69: I wonder why we don't use  [`change/2`](https://hexdocs.pm/ecto/Ecto.Changeset.html#change/2) here?

Flo0807 (approved):

lib/backpex/adapters/ecto.ex
pehbehbeh: I think it's because the function is supposed to be more of a dummy so that no errors are thrown.

If data is to be processed here, we could add that. However, it could also be a potential security risk if data is saved here in the default case.

I would say that you have to implement the function yourself if you need it.

Flo0807: I see. I thought this is a usable default.

Issue #985, opened 2025-03-20
Properly merge Tailwind CSS classes


Flo0807:
You can pass additional classes to some Backpex components. As they often come with certain defaults, we need a way to properly merge overriding and default classes. If additional classes are passed to components, these classes should have a higher priority than default classes. Currently, you sometimes need to work with `!` because default classes are of a higher priority (in terms of the place inside the final CSS file) than additional / overriding classes.

PR #1233, opened 2025-06-17
Update `input/1` component

Flo0807:
Update `Backpex.HTML.Form.input/1` to align with the *new* [Phoenix CoreComponents](https://github.com/phoenixframework/phoenix/blob/main/installer/templates/phx_web/components/core_components.ex#L170).

### DIFF
--- a/lib/backpex/html/form.ex
+++ b/lib/backpex/html/form.ex
@@ -35,14 +35,11 @@ defmodule Backpex.HTML.Form do
   attr :rest, :global, include: ~w(accept autocomplete capture cols disabled form list max maxlength min minlength
                 multiple pattern placeholder readonly required rows size step)
 
-  attr :class, :any, default: nil, doc: "additional class for the container"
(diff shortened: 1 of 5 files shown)


lib/backpex/html/form.ex
```diff
@@ -107,109 +82,84 @@ defmodule Backpex.HTML.Form do
...
+    <div class={["fieldset py-0", @class]}>
+      <label class="label cursor-pointer">
+        <input type="hidden" name={@name} value="false" />
+        <input
+          type="checkbox"
+          id={@id}
+          name={@name}
+          value="true"
+          checked={@checked}
+          class={[@input_class || "toggle toggle-primary", @errors != [] && (@error_class || "!toggle-error")]}
```
Flo0807, line 94: I don't like the use of `!` for `toggle-error`, but I guess we need a proper merge algorithm to be able to remove it. See #985

PR #1235, opened 2025-06-17
Add action items to show view

### LINKED ISSUE #872: Display Item Actions on show views
Item Actions should be displayed on show pages by default.

### Todos

- [ ] Display Item Actions on show view by default
- [ ] Handle Item Actions clicked on show view (you might need to handle some edge cases)
- [ ] Allow `:show` key to be added to `only` configuration to restrict placement
- [ ] Update docs
  - [ ] Document the new `:show` key that can be used in `only` option to configure placement (see https://hexdocs.pm/backpex/0.10.0/item-actions.html#placement-of-item-actions)

### Discussed in https://github.com/naymspace/backpex/discussions/771

<sup>Originally posted by **armanm** January 15, 2025</sup>
I saw this entry PR #591 from six month ago which pretty much the feature I'm after.

I would like to show specific item actions on the show page instead of the index. Is there a way to do this at the moment?</div>

ricmarinovic:
Closes #872.

- [x] Display Item Actions on show view by default
- [x] Handle Item Actions clicked on show view (you might need to handle some edge cases)
  - Handles the case of navigating to another page, opening a confirm dialog, opening the soft delete with a form to delete and the custom duplicate on the demo in case one wants to enable it from show.
- [x] Allow :show key to be added to only configuration to restrict placement
- [x] Update docs

Changes:
- Renamed `item` to `form_item` in the `form_component` to avoid the name clash in the `socket.assigns`.
- Changed the functions to get the `item_actions` in `resource.ex` to be more generic and allow it to be used with the new `show` parameter.

Tested manually, couldn't find a way to easily run the tests for the `demo` application.

### DIFF
--- a/lib/backpex/html/resource/resource_show.html.heex
+++ b/lib/backpex/html/resource/resource_show.html.heex
@@ -1,4 +1,26 @@
 <div>
+  <.modal
+    :if={@action_to_confirm}
+    id="action-confirm-modal"
+    title={@action_to_confirm.module.label(assigns, nil)}
+    on_cancel={JS.push("cancel-action-confirm")}
+    open={true}
+    close_label={Backpex.__("Close modal", @live_resource)}
+  >
+    <div class="px-5 py-3">
+      {@action_to_confirm.module.confirm(assigns)}
+    </div>
+    <div>
+      <.live_component
+        module={Backpex.FormComponent}
+        id={:item_action_modal}
+        live_resource={@live_resource}
+        action_type={:item}
+        {Map.drop(assigns, [:socket, :flash, :fields])}
+      />
+    </div>
+  </.modal>
+
   {@live_resource.render_resource_slot(assigns, :show, :before_page_title)}
   {@live_resource.render_resource_slot(assigns, :show, :page_title)}
 
--- a/lib/backpex/live_resource.ex
+++ b/lib/backpex/live_resource.ex
@@ -407,19 +407,21 @@ defmodule Backpex.LiveResource do
         ~H"""
         <.main_title class="flex items-center justify-between">
           {@page_title}
-          <.link
-            :if={@live_resource.can?(assigns, :edit, @item)}
-            id={"#{@live_resource.singular_name()}-edit-link"}
-            phx-hook="BackpexTooltip"
-            data-tooltip={Backpex.__("Edit", @live_resource)}
-            aria-label={Backpex.__("Edit", @live_resource)}
-            patch={Router.get_path(@socket, @live_resource, @params, :edit, @item)}
-          >
-            <Backpex.HTML.CoreComponents.icon
-              name="hero-pencil-square"
-              class="h-6 w-6 cursor-pointer transition duration-75 hover:text-primary hover:scale-110"
-            />
-          </.link>
+          <div class={["flex items-center justify-end space-x-2"]}>
+            <button
+              :for={{key, action} <- item_actions_for(:show, @item_actions)}
+              :if={@live_resource.can?(assigns, key, @item)}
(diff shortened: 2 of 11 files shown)


lib/backpex/live_resource/show.ex
```diff
@@ -70,4 +100,58 @@ defmodule Backpex.LiveResource.Show do
...
+        raise ArgumentError, """
+        Invalid return value from #{inspect(action.module)}.handle/3.
+
+        Expected: {:ok, socket}
+        Got: #{inspect(unexpected_return)}
+
+        Item Actions with no form fields must return {:ok, socket}.
+        """
+    end
+  end
```
Flo0807, line 142: We already have essentially the same code in `index.ex`. Could these functions be moved to another module (we might have to make slight adjustments to them to cover both use cases)? What do you think of `Backpex.ItemAction`?

(Credo also doesn't like the code duplication)

lib/backpex/live_resource/show.ex
```diff
@@ -39,6 +42,27 @@ defmodule Backpex.LiveResource.Show do
...
+    |> noreply()
+  end
+
+  def handle_event("cancel-action-confirm", _params, socket) do
+    socket
+    |> assign(:form_item, nil)
+    |> assign(:changeset, nil)
+    |> assign(:action_to_confirm, nil)
+    |> noreply()
+  end
```
Flo0807, line 57: We might be able to move these functions to `Backpex.ItemAction`, too (see other comment)

lib/backpex/live_resource.ex
```diff
@@ -407,19 +407,21 @@ defmodule Backpex.LiveResource do
...
+              type="button"
+              phx-click="item-action"
+              phx-value-action-key={key}
+              aria-label={action.module.label(assigns, @item)}
+              phx-hook="BackpexTooltip"
+              data-tooltip={action.module.label(assigns, @item)}
+            >
+              {action.module.icon(assigns, @item)}
+            </button>
+          </div>
```
Flo0807: I guess, we need another `div` for the tooltip that has the button inside.

It looks like there is too much space between the text and the tooltip:

image

lib/backpex/live_resource.ex
```diff
@@ -407,19 +407,21 @@ defmodule Backpex.LiveResource do
...
-            data-tooltip={Backpex.__("Edit", @live_resource)}
-            aria-label={Backpex.__("Edit", @live_resource)}
-            patch={Router.get_path(@socket, @live_resource, @params, :edit, @item)}
-          >
-            <Backpex.HTML.CoreComponents.icon
-              name="hero-pencil-square"
-              class="h-6 w-6 cursor-pointer transition duration-75 hover:text-primary hover:scale-110"
-            />
-          </.link>
+          <div class={["flex items-center justify-end space-x-2"]}>
```
Flo0807:
```suggestion
          <div class="flex items-center justify-end space-x-2">
```

lib/backpex/html/resource/resource_show.html.heex
```diff
@@ -1,4 +1,26 @@
...
+    <div>
+      <.live_component
+        module={Backpex.FormComponent}
+        id={:item_action_modal}
+        live_resource={@live_resource}
+        action_type={:item}
+        {Map.drop(assigns, [:socket, :flash, :fields])}
+      />
+    </div>
+  </.modal>
```
Flo0807, line 22: We might want to move this to another component as this is duplicated in `resource_show.html.heex` and `resource_index.html.heex`

Flo0807:
Good job and thank you for the contribution! 🙌 I've requested some changes. Also, there are some errors in the CI. @ricmarinovic

PR #1271, opened 2025-07-02
Use filtered fields in ecto adapter

krns:


lib/backpex/resource.ex
```diff
@@ -96,10 +96,9 @@ defmodule Backpex.Resource do
     {after_save_fun, opts} = Keyword.pop(opts, :after_save_fun, &{:ok, &1})
 
     adapter = live_resource.config(:adapter)
-    fields = live_resource.validated_fields()
 
     item
-    |> change(attrs, fields, assigns, live_resource, Keyword.put(opts, :action, :insert))
+    |> change(attrs, assigns.fields, assigns, live_resource, Keyword.put(opts, :action, :insert))
```
krns: I'm not sure about this one, will test in a local project

lib/backpex/live_resource/index.ex
```diff
@@ -384,14 +384,19 @@ defmodule Backpex.LiveResource.Index do
...
     socket
     |> assign(:page_title, socket.assigns.live_resource.plural_name())
+    |> assign(:fields, fields)
     |> apply_index()
     |> assign(:item, nil)
   end
 
   defp apply_action(socket, :resource_action) do
     %{live_resource: live_resource} = socket.assigns
+    fields = live_resource.validated_fields() |> LiveResource.filtered_fields_by_action(socket.assigns, :resource_action)
```
Flo0807: I wonder if we want to support `:resource_action` here. In my opinion, `resource_action` should map to `index` almost everywhere.

When I define a field with `only: [:index]`, I assume that the field will be shown when I open a resource action modal.

I never had a use case where I needed `index` and `resource_action` as two options. 

cc @pehbehbeh

lib/backpex/fields/has_many_through.ex
```diff
@@ -414,15 +414,15 @@ defmodule Backpex.Fields.HasManyThrough do
...
+  defp action_fields(fields, assigns, action), do: LiveResource.fields_by_action(fields, assigns, action)
 
   defp assign_fallback_child_fields(assigns) do
     case Map.has_key?(assigns.field_options, :child_fields) do
       true ->
         assigns
 
       false ->
-        fields = assigns.field_options.live_resource.validated_fields()
+        fields = assigns.field_options.live_resource.fields(:index, assigns)
```
krns: I'm not sure about this one.. the fallback child fields will not work because the assigns are the assigns of the field.

lib/backpex/fields/has_many_through.ex
```diff
@@ -414,15 +414,15 @@ defmodule Backpex.Fields.HasManyThrough do
     {:noreply, socket}
   end
 
-  defp action_fields(fields, assigns, action), do: LiveResource.filtered_fields_by_action(fields, assigns, action)
+  defp action_fields(fields, assigns, action), do: LiveResource.fields_by_action(fields, assigns, action)
```
krns: What is `action` here? I think `fields_by_action` should replaced too

Flo0807: Basically, it is the `live_action` to filter the list of child fields and support `only` / `except` options. Same like on index.

lib/backpex/live_resource/form.ex
```diff
@@ -16,9 +16,9 @@ defmodule Backpex.LiveResource.Form do
     |> assign(:live_resource, live_resource)
     |> assign(:panels, live_resource.panels())
     |> assign(:fluid?, live_resource.config(:fluid?))
+    |> assign(:fields, live_resource.fields(live_action, socket.assigns))
```
krns, line 19: The fields are still in the assigns, but only to render them in the templates via `@fields` (see also `index.ex` and `show.ex`).

lib/backpex/live_resource/index.ex
krns: > I never had a use case where I needed index and resource_action as two options.

Same for me

lib/backpex/field.ex
```diff
@@ -427,6 +427,7 @@ defmodule Backpex.Field do
       end
     ]
 
+    fields = live_resource.fields(:index, assigns)
     result = Backpex.Resource.update(item, change, fields, assigns, live_resource, opts)
```
Flo0807: As we have access to the `live_resource`, can we call `live_resource.fields` from within the `update` / adapter function? We already do this in `list_query`, too.

krns: Yes, now that we have `live_resource.fields`, I'll do that.

krns: See https://github.com/naymspace/backpex/pull/1271#pullrequestreview-2978135109

krns: 939dc36a82a6400445b2498b741242d6b041f5c6

lib/backpex/adapters/ecto.ex
```diff
@@ -75,10 +75,10 @@ defmodule Backpex.Adapters.Ecto do
   Gets a database record with the given primary key value.
   """
   @impl Backpex.Adapter
-  def get(primary_value, assigns, live_resource) do
+  def get(primary_value, assigns, fields, live_resource) do
```
krns: I added the `fields` parameter to this function, because we call it (via `Resource.get`) in `form.ex`, `index.ex` and `show.ex` and need the `fields` explicitly or alternatively the `live_action` and `live_resource` to get the list of fields.

What bothers me is that all other functions do not need the `fields`, because they are not used in different places and can hard code the `live_action` inside the adapter function (see `Backpex.Adapters.Ecto.list/3`).

The `live_action` should also be inside the assigns, but using the assigns felt like a step backwards.

lib/backpex/fields/has_many_through.ex
```diff
@@ -414,15 +408,24 @@ defmodule Backpex.Fields.HasManyThrough do
     {:noreply, socket}
   end
 
-  defp action_fields(fields, assigns, action), do: LiveResource.filtered_fields_by_action(fields, assigns, action)
+  defp action_fields(fields, :resource_action), do: action_fields(fields, :index)
```
krns: I'm not quite sure if this change makes sense. Has the HasManyThrough field ever worked in a resource action?

guides/upgrading/v0.14.md
```diff
@@ -81,3 +81,12 @@ def changeset(change, attrs, _metadata) do
   |> Ecto.Changeset.validate_required([:field1])
 end
 ```
+
+## Resource.update/6 has been changed
```
Flo0807: As we released v0.14 this has to be moved to v0.15 upgrade guide.

guides/live_resource/navigation.md
```diff
@@ -25,7 +25,6 @@ Backpex supports the following live actions:
 - `:new` - The form for creating a new resource
 - `:edit` - The form for editing an existing resource
 - `:show` - The detailed view of a single resource
-- `:resource_action` - An open resource action on the list view
```
Flo0807: `:resource_action` live action is still passed to `return_to`, right?

guides/upgrading/v0.14.md
```diff
@@ -81,3 +81,12 @@ def changeset(change, attrs, _metadata) do
...
+
+We no longer pass the `fields` to `Resource.update/6`. Therefore, the arity of this functions has changed to [`Resource.update/5`](). This aligns the function `insert` and `update` of the Resource module.
```
Flo0807:
```suggestion
We no longer pass the `fields` to `Resource.update/6`. Therefore, the arity of this functions has changed to `Backpex.Resource.update/5`. This aligns the function `insert` and `update` of the Resource module.
```

guides/upgrading/v0.7.md
```diff
@@ -39,8 +39,8 @@ end
 We have updated certain functions in `Backpex.Resource`.
 
 The following functions are affected:
-- `Backpex.Resource.update/6` (`update/5` before)
-- [`Backpex.Resource.insert/6`]() (`insert/5` before)
+- `update/6` (`update/5` before)
+- `insert/6` (`insert/5` before)
 - [`Backpex.Resource.change/7`]()
 - [`Backpex.Resource.put_assocs/2`]() (has been removed)
```
Flo0807:
```suggestion
- `update/6` (`update/5` before)
- `insert/6` (`insert/5` before)
- `change/7`
- `put_assocs/2` (has been removed)
```

lib/backpex/adapter.ex
```diff
@@ -27,7 +27,7 @@ defmodule Backpex.Adapter do
 
   Should return `nil` if no result was found.
   """
-  @callback get(term(), map(), module()) :: {:ok, struct() | nil} | {:error, term()}
+  @callback get(term(), list(), module()) :: {:ok, struct() | nil} | {:error, term()}
```
Flo0807: `assigns` is a map?

lib/backpex/fields/has_many_through.ex
```diff
@@ -414,15 +408,24 @@ defmodule Backpex.Fields.HasManyThrough do
...
     case Map.has_key?(assigns.field_options, :child_fields) do
       true ->
         assigns
 
       false ->
-        fields = assigns.field_options.live_resource.validated_fields()
+        # It's currently not supported to add a `can?` options to the child fields. Therefore we
+        # are passing an empty map as the assigns.
+        # See https://github.com/naymspace/backpex/pull/1271
+        fields = assigns.field_options.live_resource.fields(:index, %{})
```
Flo0807, line 426: Do we remove this feature with this MR? Was ist possible before?

lib/backpex/live_resource/index.ex
```diff
@@ -537,14 +535,15 @@ defmodule Backpex.LiveResource.Index do
     %{
       live_resource: live_resource,
       params: params,
-      fields: fields,
```
Flo0807, line 556: Isn't this still possible as `mount` assigns fields?

lib/backpex/live_resource/index.ex
```diff
@@ -317,9 +318,8 @@ defmodule Backpex.LiveResource.Index do
   end
 
   defp assign_active_fields(socket, session) do
-    fields =
-      socket.assigns.live_resource.validated_fields()
-      |> LiveResource.filtered_fields_by_action(socket.assigns, :index)
+    %{live_resource: live_resource} = socket.assigns
+    fields = live_resource.fields(:index, socket.assigns)
```
Flo0807:
```suggestion
    fields = socket.assigns.live_resource.fields(:index, socket.assigns)
```

lib/backpex/live_resource/index.ex
```diff
@@ -566,14 +565,14 @@ defmodule Backpex.LiveResource.Index do
   defp maybe_assign_metrics(socket) do
     %{
       live_resource: live_resource,
-      fields: fields,
```
Flo0807, line 585: Isn't this still possible as `mount` assigns fields?

lib/backpex/live_resource/index.ex
```diff
@@ -431,6 +428,8 @@ defmodule Backpex.LiveResource.Index do
 
     schema = live_resource.adapter_config(:schema)
 
+    fields = live_resource.fields(:index, socket.assigns)
```
Flo0807: Can we fetch `fields` from the `assigns` here as they were already assigned in `mount`?

lib/backpex/live_resource/form.ex
```diff
@@ -91,17 +91,10 @@ defmodule Backpex.LiveResource.Form do
...
-      |> LiveResource.filtered_fields_by_action(socket.assigns, live_action)
-
-    assign(socket, :fields, fields)
-  end
-
   defp assign_changeset(socket, live_action) do
-    %{live_resource: live_resource, item: item, fields: fields} = socket.assigns
+    %{live_resource: live_resource, item: item} = socket.assigns
     changeset_fun = changeset_fun(live_resource, live_action)
+    fields = live_resource.fields(live_action, socket.assigns)
```
Flo0807: We already have `fields` in the list of assigns, so why do we need to call this function again?

Flo0807 (changes requested):
We call `fields/` in so many places. Can't we just assign fields once and then reuse them from the `assigns`?

lib/backpex/fields/has_many_through.ex
Flo0807: I don't think so, but I'd still keep your change.

Flo0807 (approved):

PR #1289, opened 2025-07-11
Fix filter docs

krns:


Flo0807:
We need to change it in the custom filter explanation as well:

> The `query/3` function filters the data based on the selected value.

In the custom filter (`render/1` function) we call:

```
assigns = assign(assigns, :label, option_value_to_label(options(), assigns.value))
```

I guess we need to call `my_options/0` because the filter does not define an `options/0` function 🤔 We might want to change this in this MR, too.

@krns

Flo0807 (approved):

PR #1307, opened 2025-07-18
Use masked input for `Backpex.Fields.Currency`

### LINKED ISSUE #1280: `Money` dependency conflicts with `ExMoney` dependency in project
**Describe the bug**
When I add Backpex to my project that depends on `ExMoney`, I receive the following compiler error half the times when compiling the project:

```
Unchecked dependencies for environment test:
* ex_money (Hex package)
  could not find an app file at "_build/test/lib/ex_money/ebin/ex_money.app". This may happen if the dependency was not yet compiled or the dependency indeed has no app file (then you can pass app: false as option)
** (Mix) Can't continue due to errors on dependencies
```

This probably happens because Backpex depends on `money` which for unfortunate reasons conflicts with `ex_money` which uses the same `Money.*` namespace.

**To Reproduce**
1. Add Backpex to a project that depends also on `ex_money`.
2. Compile the project a few times

**Expected behavior**
Money and ExMoney would not share the same `Money.*` namespace but that ship has sailed.

Maybe Backpex could instead make the money dependency optional and offer the `Currency` field only if the Money library is loaded? Ideally, it would support both libraries of course, but that'd be difficult since now you also use Money for storing currencies in the database through the `amount_type.ex`. Did you know that ExMoney offers such Ecto types out-of-the-box by the way? ;)

Maybe this doesn't have a solution though 🤷 I just wanted to put this out there.

**Screenshots**

**Please complete the following information:**
 - OS: macOS
 - Elixir Version: 1.18.4
 - Backpex Version: 0.13.0

**Additional context**

Flo0807:
closes #1280


lib/backpex/fields/currency.ex
```diff
@@ -91,25 +111,7 @@ defmodule Backpex.Fields.Currency do
   def search_condition(schema_name, field_name, search_string) do
```
Flo0807, line 117: Still, the UX when searching for currency is not very good, as you have to search for integer values. We could add another `field_option` that is used in this function to convert the search string into an integer. (However, we would need to modify the function to receive the field options...)

WDYT? @krns @pehbehbeh

assets/js/hooks/_currency_input.js
```diff
@@ -0,0 +1,49 @@
...
+  },
+  handleMaskChange () {
+    this.hiddenInput.value = this.rawValue(this.mask.value)
+  },
+  destroyed () {
+    this.mask.destroy()
+  },
+  rawValue (value) {
+    return value
+      .replace(this.el.dataset.unit || '', '')
```
krns, line 45: Maybe the `.` is a better default?

lib/backpex/html/form.ex
```diff
@@ -164,6 +164,90 @@ defmodule Backpex.HTML.Form do
     """
   end
 
+  @doc """
+  Renders a masked input.
```
krns:
```suggestion
  Renders a masked input for currencies.
```

krns (commented):
Overall, I only have one question about the default values.

Have you been able to test the changes of #1359 yet?

assets/js/hooks/_currency_input.js
Flo0807: This is about replacing the unit (not the delimiter).

krns: Oh yes.. but the questions remains^^

Flo0807: Why do we want to replace the unit with `.`?

Flo0807: If there is no unit, we don't want to replace anything, so why do we want to use `.` as a default?

Flo0807: I added this line to remove the unit from the currency string, because ` €` is invalid and results in a `is invalid` changeset error instead of `can't be blank`

lib/backpex/fields/currency.ex
krns: Does the same apply for Filters like a Range filter?

To have some kind of options to transform the search term sounds good.

assets/js/hooks/_currency_input.js
krns: Sorry for the confusion, I will add a new comment in the right place.

lib/backpex/fields/currency.ex
```diff
@@ -7,6 +7,28 @@ defmodule Backpex.Fields.Currency do
...
+      type: :string,
+      default: ","
+    ],
+    thousands_separator: [
+      doc:
+        "Character used as the thousands separator, e.g. '.' or ','. Make sure this value matches the one you've configured in your Money library.",
+      type: :string,
+      default: "."
     ]
   ]
```
krns, line 38: Should we use more common defaults like `$ 1,000.00`?

lib/backpex/fields/currency.ex
Flo0807: We have the same problem when using a range filter. The `search_condition` is not used there, so I guess we should leave it like it is and find a solution that covers both filters and field search.

guides/upgrading/v0.16.md
```diff
@@ -0,0 +1,74 @@
...
+  default_currency: :EUR,
+  separator: ".",
+  delimiter: ",",
+  symbol_on_right: true,
+  symbol_space: true
+```
+
+The default values should ideally match the field options from step 1.
+
+3. Replace [`Backpex.Ecto.AmountType`]() with [`Backpex.Ecto.AmountType`]()
```
pehbehbeh:
```suggestion
3. Replace [`Backpex.Ecto.AmountType`]() with [`Money.Ecto.Amount.Type`]()
```

guides/upgrading/v0.16.md
```diff
@@ -12,6 +12,67 @@ defp deps do
...
+
+2. Configure defaults for `money` in your `config/config.exs`
+
+```elixir
+config :money,
+  default_currency: :EUR,
+  separator: ".",
+  delimiter: ",",
+  symbol_on_right: true,
+  symbol_space: true
```
krns, line 52: Align config keys:

- separator vs. thousands_separator
- delimiter vs. radix 
- symbol_on_right vs. unit_position
- symbol_space not available on field level?

krns: Ahh, got it.. thats the money config vs. backpex config.. hmhmh

PR #1308, opened 2025-07-18
Require `changeset/3` in ItemActions

Flo0807:


lib/backpex/item_actions/item_action.ex
```diff
@@ -132,6 +125,37 @@ defmodule Backpex.ItemAction do
...
+          def changeset(change, attrs, _metadata) do
+            change
+            |> Ecto.Changeset.cast(attrs, [:field1, :field2])
+            |> Ecto.Changeset.validate_required([:field1])
+          end
+          """
+      end
+    rescue
+      UndefinedFunctionError ->
+        # fields/0 is not defined, which means it will use the default empty list
```
Flo0807: We might want to remove the fallback `fields/0` function.

If the there is a `fields/0` function -> call it and check for non empty list
If there is **no** `fields/0` function -> we are fine because item action has no form anyway

No need to implement a `fields/0` fallback...

Edit: Removing the `fields/0` function can result in some refactoring, e.g., we always iterate over the fields if we show the confirmation modal.

guides/actions/item-actions.md
```diff
@@ -169,36 +176,33 @@ defmodule DemoWeb.ItemAction.SoftDelete do
...
 
-The `c:Backpex.ItemAction.handle/3` function is called when the item action is triggered. The handle function receives the socket, the items that should be affected by the action, and the parameters that were submitted by the user.
-
-By default an item action is triggered immediately when the user clicks on the corresponding icon in the resource table or in the show view, but an item actions also supports a confirmation dialog. To enable the confirmation dialog you need to implement the `c:Backpex.ItemAction.confirm/1` function and return a string that will be displayed in the confirmation dialog. The confirmation dialog will be displayed when the user clicks on the icon in the resource table.
-
-You might want to use the `c:Backpex.ItemAction.cancel_label/1` (defaults to "Cancel") and `c:Backpex.ItemAction.confirm_label/1` (defaults to "Apply") functions to set the labels of the buttons in the dialog.
+> #### Important {: .note}
+> If your ItemAction has form fields, you must also implement the `c:Backpex.ItemAction.confirm/1` function.
```
Flo0807, line 208: I don't like that you have to implement `confirm/1` if your ItemAction has form inputs. In case you just want to show the inputs with out any confirmation text. WDYT?

Flo0807: But changing this behavior does not need to be part of this MR...

krns: 👍

PR #1322, opened 2025-07-22
Install quokka and format code

Flo0807:


demo/.formatter.exs
```diff
@@ -1,7 +1,12 @@
...
-  plugins: [Spark.Formatter, TailwindFormatter, Phoenix.LiveView.HTMLFormatter],
+  plugins: [Spark.Formatter, TailwindFormatter, Phoenix.LiveView.HTMLFormatter, Quokka],
   inputs: ["*.{heex,ex,exs}", "priv/*/seeds.exs", "{config,lib,test}/**/*.{heex,ex,exs}"],
-  subdirectories: ["priv/*/migrations"]
+  subdirectories: ["priv/*/migrations"],
+  quokka: [
+    exclude: [
+      :module_directives
+    ]
+  ]
```
Flo0807, line 11: I excluded `module_directives` because I don't like that almost every nested module is then aliased on top of the module.

Flo0807:
See https://github.com/naymspace/backpex/pull/1464

PR #1353, opened 2025-07-31
Fix index edit not shown behind resource action modal

Flo0807:
If the index view had editable fields, the layout would shift when the resource action modal was opened because the inputs were only shown for `:index` `live_action`, not `:resource_action`.

See the upload action in the latest demo: https://demo.backpex.live/admin/users


Flo0807:
If we treat `:index` and `:resource_action` as the same actions, these types of bugs would not occur.

cc @krns @phyr97

PR #1359, opened 2025-08-01
Bundle JS files

Flo0807:


Flo0807:
@pehbehbeh I decided to go for an up-to-date check in the pipeline instead of auto commiting the bundle files. This simplifies the release.

PR #1361, opened 2025-08-03
Fix entries were excluded from index view under certain conditions

### LINKED ISSUE #562: searchable: true + NULL value causes the record not to be displayed
**Describe the bug**
I have a nullable string field, with `searchable: true`. When the field is null, it will not be displayed in the index even without touching the search field. When I remove searchable: true or insert an value, it will be displayed in the index table.

**To Reproduce**
- Create a record with a nullable string field
- Set it to searchable
- Create a record and leave the field on NULL / nil
- Visit index page
- Record is not displayed

**Expected behavior**
I would have expected to see the record on index page unless I start a search.

**Please complete the following information:**
 - OS: MacOS
 - Browser Arc
 - Elixir Version 1.17.2
 - Backpex Version 0.6.0

Flo0807:
closes #562 

We did not skip applying the search conditions if the search string was empty (`""`).

This resulted in conditions such as `ilike(a0.city, "%%")`being applied. While this was OK for string values, it excluded `NULL` values.

The solution is to skip applying search conditions when the search string is empty.

The fix can be tested by making the `address.city` field nullable and disabling search for all `adress` fields except for the `city` field.

### DIFF
--- a/lib/backpex/adapters/ecto.ex
+++ b/lib/backpex/adapters/ecto.ex
@@ -135,6 +135,8 @@ defmodule Backpex.Adapters.Ecto do
 
   def apply_search(query, _schema, nil, {_search_string, []}), do: query
 
+  def apply_search(query, _schema, nil, {"", _searchable_fields}), do: query
+
   def apply_search(query, _schema, nil, {search_string, searchable_fields}) do
     search_string = "%#{search_string}%"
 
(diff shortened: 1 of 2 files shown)


Flo0807:
> Maybe we should add tests for that.

I thought the same, but without a way to set up LiveResources for testing, it becomes complicated as the bug only occurs if the field is the only searchable field in the resource. Therefore, we might have to update the demo in order to test it.

Alternatively, we could test the list function from the Ecto adapter.

PR #1435, opened 2025-08-28
Support Phoenix 1.8

Flo0807:


lib/backpex/html/layout.ex
```diff
@@ -8,6 +8,42 @@ defmodule Backpex.HTML.Layout do
...
+
+  def layout(assigns) do
+    case assigns.live_resource.config(:layout) do
+      {module, fun} ->
+        apply(module, fun, [assigns])
+
+      fun when is_function(fun, 1) ->
+        fun.(assigns)
+
+      wrong_config ->
```
Flo0807: I noticed that we probably don't need this case as the option is validated via NimbleOptions, right?

mix.exs
```diff
@@ -53,7 +53,7 @@ defmodule Backpex.MixProject do
       {:money, "~> 1.13"},
 
       # phoenix
-      {:phoenix, "~> 1.7.6"},
+      {:phoenix, "~> 1.7"},
```
Flo0807: It is still possible to use Phoenix 1.7

demo/config/dev.exs
```diff
@@ -24,9 +24,7 @@ config :demo, DemoWeb.Endpoint,
     patterns: [
       ~r"demo/priv/static/.*(js|css|png|jpeg|jpg|gif|svg)$",
       ~r"demo/priv/gettext/.*(po)$",
-      ~r"demo/lib/demo_web/helpers.ex$",
-      ~r"demo/lib/demo_web/(live|views)/.*(ex)$",
-      ~r"demo/lib/demo_web/templates/.*(eex)$",
+      ~r"lib/demo_web/(?:controllers|live|components|router)/?.*\.(ex|heex)$",
```
pehbehbeh, line 27: Does this still work?

mix.exs
pehbehbeh: But we cannot be sure that it is possible to use Phoenix 1.9

lib/backpex/html/layout.ex
pehbehbeh: Yes

demo/mix.exs
```diff
@@ -97,7 +104,8 @@ defmodule Demo.MixProject do
         "esbuild backpex --minify",
         "esbuild default --minify",
         "phx.digest"
-      ]
+      ],
+      precommit: ["compile --warning-as-errors", "deps.unlock --unused", "format", "test"]
```
pehbehbeh: Do we use this anywhere?

demo/mix.exs
```diff
@@ -12,7 +12,14 @@ defmodule Demo.MixProject do
...
       aliases: aliases(),
-      gettext: gettext()
+      gettext: gettext(),
+      listeners: [Phoenix.CodeReloader]
+    ]
+  end
+
+  def cli do
+    [
+      preferred_envs: [precommit: :test]
```
pehbehbeh: Can be removed? (see below)

mix.exs
Flo0807: Updated to `">= 1.7.6 and < 1.9.0"`

demo/mix.exs
Flo0807: Yes, removed it.

demo/config/dev.exs
Flo0807: yes

PR #1437, opened 2025-08-29
fix: make form submit error pattern match include Ash.Changeset

enoonan:
Previously when saving a form with `AshBackpex`, if there were errors in the form it would crash (and reload). 

This was because the error handling pattern match only looked for `Ecto.Changeset`. 

This PR just adds `Ash.Changeset` to the pattern match.

### DIFF
--- a/lib/backpex/live_components/form_component.ex
+++ b/lib/backpex/live_components/form_component.ex
@@ -236,7 +236,7 @@ defmodule Backpex.FormComponent do
         |> push_navigate(to: return_to)
         |> noreply()
 
-      {:error, %Ecto.Changeset{} = changeset} ->
+      {:error, %struct{} = changeset} when struct in [Ecto.Changeset, Ash.Changeset] ->
         form = Phoenix.Component.to_form(changeset, as: :change)
 
         send(self(), {:update_changeset, changeset})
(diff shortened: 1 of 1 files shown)


Flo0807:
Hey, thanks for the PR!

Can we somehow solve this via the adapter only? The goal should be to have everything Ash-related in the adapter module.

@enoonan

Flo0807:
Hey @enoonan, I wondered if it is possible to pattern match on struct type? Would that also work for you?

```elixir
{:error, %Ecto.Changeset{} = changeset} when is_struct(changeset) ->
```

enoonan:
Just made that change! 

Another potential option would be creating a custom guard that checks whether `changeset` implements `Phoenix.HTML.FormData`. That would accommodate any future adapters as well: 

https://hexdocs.pm/elixir/1.12.3/Kernel.html#defguard/1
https://hexdocs.pm/elixir/Protocol.html#assert_impl!/2

lib/backpex/live_components/form_component.ex
```diff
@@ -279,7 +279,7 @@ defmodule Backpex.FormComponent do
         |> push_navigate(to: return_to)
         |> noreply()
 
-      {:error, %Ecto.Changeset{} = changeset} ->
+      {:error, changeset} when is_struct(changeset)  ->
```
Flo0807:
```suggestion
      {:error, changeset} when is_struct(changeset) ->
```

Flo0807 (approved):

Flo0807:
Thanks for the PR! 🙌

PR #1466, opened 2025-09-05
Simplify query function in MultiSelect filter

gmazzamuto:
Hi all,

it seems the `query` function in `Backpex.Filters.MultiSelect` is overly complicated, and can be rewritten in just a one-liner. I have also renamed the argument `value` to `values` for better coherence.

Hope this helps!


Flo0807:
Thank you! 🙌

PR #1470, opened 2025-09-06
Update package.json

nolman:
I followed the installation guide and I would get an error

```
 [WARNING] Import "Hooks" will always be undefined because the file "../deps/backpex/priv/static/js/backpex.cjs.js" has no exports [import-is-undefined]

    js/app.js:27:9:
      27 │ import { Hooks as BackpexHooks } from "backpex";
 ```
 
 Updating this resolved the error, not sure if this is the right way to solve it tho

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.14.0",
   "license": "MIT",
   "type": "module",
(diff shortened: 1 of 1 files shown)


Flo0807:
Hey, can you share your esbuild / bundler configuration from `config/config.exs`?

PR #1472, opened 2025-09-06
Add `Backpex.Fields.Email`

nolman:
I wanted a field that had a proper email input on mobile - not sure if you had a different way you wanted to do this


Flo0807:
Thanks for the contribution! 🚀

PR #1481, opened 2025-09-08
Make igniter and igniter_js optional dependencies

Flo0807:


demo/mix.exs
```diff
@@ -70,6 +70,8 @@ defmodule Demo.MixProject do
       # application
       {:backpex, path: "../."},
       {:phoenix_ecto, "~> 4.0"},
+      {:igniter, "~> 0.6", optional: true},
+      {:igniter_js, "~> 0.4", optional: true},
```
Flo0807:
```suggestion
      {:igniter, "~> 0.6"},
      {:igniter_js, "~> 0.4"},
```

I guess we don't need to specify `optional` in the demo `mix.exs`.

PR #1511, opened 2025-09-15
doc fix Change layout rendering to use inner_block

bryanfaria:
Fix inconsistent documentation. 
Updated the layout rendering to use inner_block instead of inner_content.


Flo0807:
Thanks for the fix! 🙌

PR #1528, opened 2025-09-18
Add support for option groups in `Select` and `MultiSelect` fields

gmazzamuto:
Hi all! First of all my congrats for Backpex!

Here is a PR with some enhancements:
* the first commit is the simplest, it just replaces a complicated `Enum.reduce` followed by `Enum.reverse` with a single call to `Enum.filter`
* the other commits add support for option groups in the `Select` and `MultiSelect` fields. Option groups are already supported by Phoenix via [Phoenix.HTML.Form.options_for_select/2](https://hexdocs.pm/phoenix_html/Phoenix.HTML.Form.html#options_for_select/2), but don't work out of the box in Backpex. The changes for `Select` are a simple fix, the implementation for `MultiSelect` is just a bit more complicated. See below for some screenshots.

I have made other small improvements if you are interested!

Option groups in `Select` (as seen in Firefox):
* Screenshot_20250918_225026

Option groups in `MultiSelect`:
* Screenshot_20250918_225138

### DIFF
--- a/lib/backpex/html/form.ex
+++ b/lib/backpex/html/form.ex
@@ -442,6 +442,7 @@ defmodule Backpex.HTML.Form do
       <div class="ml-4">
         <.multi_select_option
           :for={{lab, val} <- @value}
+          class="mt-2"
           label={lab}
           value={val}
           event_target={@event_target}


gmazzamuto:
Hi there! I just wanted to let you know that I have updated these changes to reflect the recent addition of the `dropdown` component.

If interested, let me know if you would like me to split this PR in two, one for `Select` and one for `MultiSelect`. Btw, out of 4 commits, the first 3 are very small patches, only the last one is a bit more complicated.

Flo0807:
> Hi there! I just wanted to let you know that I have updated these changes to reflect the recent addition of the `dropdown` component.
> 
> If interested, let me know if you would like me to split this PR in two, one for `Select` and one for `MultiSelect`. Btw, out of 4 commits, the first 3 are very small patches, only the last one is a bit more complicated.

Thanks! I am going to look into this pull request this week.

demo/lib/demo_web/live/user_live.ex
```diff
@@ -196,7 +196,11 @@ defmodule DemoWeb.UserLive do
       permissions: %{
         module: Backpex.Fields.MultiSelect,
         label: "Permissions",
-        options: fn _assigns -> [{"Delete", "delete"}, {"Edit", "edit"}, {"Show", "show"}] end
+        options: [
+          {"Foo", "foo"},
```
Flo0807:
```suggestion
```

We should remove this option or turn it into something more realistic.

lib/backpex/fields/multi_select.ex
```diff
@@ -66,50 +66,64 @@ defmodule Backpex.Fields.MultiSelect do
...
+        case PhoenixForm.input_value(assigns.form, name) do
+          value when is_binary(value) -> [value]
+          value when is_list(value) -> value
+          _value -> []
+        end
       else
-        value = Map.get(item, name)
-
-        if value, do: value, else: []
+        Map.get(item, name, []) || []
```
Flo0807:
```suggestion
        Map.get(item, name, [])
```

lib/backpex/fields/multi_select.ex
```diff
@@ -207,27 +219,35 @@ defmodule Backpex.Fields.MultiSelect do
...
+    filtered =
+      Enum.filter(value, fn {_label, _value} -> String.downcase(label) |> String.contains?(search_input_downcase) end)
+
+    if not Enum.empty?(filtered), do: {label, filtered}
+  end
+
+  defp filter_option(label, _value, search_input) do
+    search_input_downcase = String.downcase(search_input)
+
+    String.downcase(label) |> String.contains?(search_input_downcase)
```
Flo0807:
```suggestion
    label
    |> String.downcase() 
    |> String.contains?(search_input_downcase)
```

Flo0807 (changes requested):

lib/backpex/fields/multi_select.ex
gmazzamuto: Unfortunately this is needed, because `Map.get(item, name, [])` can still return `nil`, whereas we want to always return a list.
For example:
```elixir
item = %{name: nil}
Map.get(item, :name, [])
```
will return `nil`.

Maybe it's better to rewrite it like this:
```elixir
Map.get(item, name) || []
```

Flo0807: I see. Let's stick with `Map.get(item, name) || []` 🙂

gmazzamuto:
you're right! it should be fixed now

gmazzamuto:
Awesome! thanks for your review! Indeed I have a couple more improvements that I was holding back, waiting for these changes to be merged first :smile:

PR #1544, opened 2025-09-23
Fix picker appearance for select input

Flo0807:


lib/backpex/html/form.ex
```diff
@@ -111,7 +111,9 @@ defmodule Backpex.HTML.Form do
           name={@name}
           class={[
             @input_class || "select w-full",
-            @errors != [] && (@error_class || "select-error text-error-content bg-error/10")
+            @errors != [] &&
+              (@error_class ||
+                 "select-error text-error-content bg-error/10 [&.select::picker(select)]:bg-base-100 [&.select::picker(select)]:text-base-content")
```
Flo0807, line 116: It works when you target the select class first @pehbehbeh

PR #1553, opened 2025-09-25
Improve filter components

Flo0807:


demo/lib/demo_web/live/product_live.ex
```diff
@@ -21,7 +21,7 @@ defmodule DemoWeb.ProductLive do
     [
       quantity: %{
         module: DemoWeb.Filters.ProductQuantityRange,
-        label: "Quantity"
+        label: "QTY"
```
Flo0807, line 24: Unfortunately, this label option did not work. While it was displayed in the dropdown, it was not inside filter badges. fixed with this PR.

lib/backpex/html/core_components.ex
```diff
@@ -20,40 +20,4 @@ defmodule Backpex.HTML.CoreComponents do
...
-  @doc type: :component
-
-  attr :clear_event, :string, default: "clear-filter", doc: "event name for removing the badge"
-  attr :filter_name, :string, required: true
-  attr :label, :string, required: true
-  attr :live_resource, :atom, default: nil
-
-  slot :inner_block
-
-  def filter_badge(assigns) do
```
Flo0807, line 36: IMO it makes sense to have all filter components in one place. That's why I moved `filter_badge` from `core_components` to `html/resource.ex`. I even thought about creating a dedicated module for them, as we have 5 filter components...

lib/backpex/html/resource.ex
```diff
@@ -198,116 +199,180 @@ defmodule Backpex.HTML.Resource do
...
+  def index_filter(assigns) do
+    assigns = assign(assigns, :filter_count, Enum.count(assigns.filter_options))
+
+    ~H"""
+    <.index_filter_dropdown :if={@filters != []} live_resource={@live_resource} filter_count={@filter_count}>
+      <.index_filter_forms filters={@filters} filter_options={@filter_options} live_resource={@live_resource} {assigns} />
+    </.index_filter_dropdown>
+    <%= for {key, value} <- @filter_options do %>
+      <% filter = Keyword.get(@filters, String.to_existing_atom(key)) %>
+      <% label = Map.get(filter, :label, filter.module.label()) %>
```
Flo0807: bad practice?

Flo0807:
With the refactoring, you should be able to reuse the filter components more easily (e.g. on custom views)

For example:

```heex
<.filter_dropdown filter_count={@filter_count}>
  <.form :let={f} for={@form} phx-change="change-filter" class="space-y-5">
    <.filter_form_field filter_name={:status} label="Status" show_clear_button={@filter_opts[:status]}>
      <.input type="select" prompt="Select status..." options={@status_options} />
    </.filter_form_field>
  </.form>
</.filter_dropdown>
```

lib/backpex/html/resource.ex
pehbehbeh: Yeah, there are certainly nicer ways to solve this... we could assemble the list with everything already in it as an assign?

pehbehbeh (approved):

PR #1594, opened 2025-10-13
Fix a11y issues

Flo0807:
- Remove `continue-on-error: true` from CI
- Fix all a11y issues from browser tests


lib/backpex/html/core_components.ex
```diff
@@ -27,7 +27,7 @@ defmodule Backpex.HTML.CoreComponents do
   ## Examples
 
       <.dropdown id="user-menu">
-        <:trigger class="btn btn-primary btn-sm">
+        <:trigger class="btn btn-primary btn-sm" aria-label="User Menu">
           User Menu
```
SebastianMueller87, line 31: We do not use translation in here? 🤔

lib/backpex/html/core_components.ex
```diff
@@ -63,18 +64,25 @@ defmodule Backpex.HTML.CoreComponents do
...
-      <div id={"#{@id}-trigger"} role="button" tabindex="0" aria-haspopup="menu" class={@trigger && @trigger[:class]}>
+    <div id={@id} class={["dropdown", @class]}>
+      <div
+        id={"#{@id}-trigger"}
+        role="button"
+        tabindex="0"
+        aria-haspopup="true"
+        aria-label={@trigger && @trigger[:aria_label]}
+        class={@trigger && @trigger[:class]}
+        {@rest}
```
SebastianMueller87: Its correct, @rest is now applied on the inner div?

lib/backpex/html/core_components.ex
Flo0807: This is intentional because it is example code on how to use the component.

lib/backpex/html/core_components.ex
Flo0807: No, I guess it should be applied to the outer div. I am going to fix this.

lib/backpex/html/core_components.ex
SebastianMueller87: Uhh, missed that. Embarrasing. :-D

PR #1607, opened 2025-10-18
added `embed_one` type to the Backpex.Fields.InlineCRUD

thanos:
As discussed in #1583, the pull request adds the handling of `embeds_one` fields with ` Backpex.Fields.InlineCRUD`. Also added an example field to the `Demo.Product` and its corresponding admin `live_reasource`.


thanos:
@Flo0807 - this is good to go - apologies for the delay.

PR #1616, opened 2025-10-22
this pull request adds the handling of map fields with Backpex.Fields.InlineCRUD.

thanos:
This is can really useful:
  Controls what values are stored in a map
  Allows you to simulate Polymorphism as illustrated by example use case I added to Demo.
  I know the example could look too complicated at first, and it's your library so it's your call , but I think it's fun and it is a good example of the power of Backpex!

All the best to your team  - @thanos


thanos:
Sorry I missed this - I'll follow through.

PR #1621, opened 2025-10-23
Add support for most field types in `Backpex.Fields.InlineCRUD`

gmazzamuto:
Hi all!

These relatively simple changes make it possible to use almost all Fields with `Backpex.Fields.InlineCRUD`.

Hope you'll like it! :)

Screenshot 2025-10-24 at 00-19-15 Edit Product · Backpex
Screenshot 2025-10-24 at 00-18-52 Product · Backpex

### DIFF
--- a/lib/backpex/fields/inline_crud.ex
+++ b/lib/backpex/fields/inline_crud.ex
@@ -7,8 +7,12 @@ defmodule Backpex.Fields.InlineCRUD do
     ],
     child_fields: [
       doc: """
-      A list of child input fields. All field types except `Backpex.Fields.HasManyThrough`, `Backpex.Fields.InlineCRUD`
-      and `Backpex.Fields.Upload` are supported.
+      A list of child input fields.
+
+      The following fields are **not** supported:
+      - `Backpex.Fields.HasManyThrough`
+      - `Backpex.Fields.InlineCRUD`
+      - `Backpex.Fields.Upload`
 
       You can add additional classes to child field inputs by setting the class option in the list of `child_fields`.
       The class can be a string or a function that takes the assigns and must return a string.


demo/lib/demo_web/live/product_live.ex
```diff
@@ -96,8 +96,26 @@ defmodule DemoWeb.ProductLive do
...
           },
           url: %{
-            module: Backpex.Fields.Text,
+            module: Backpex.Fields.URL,
             label: "URL"
+          },
+          country: %{
+            module: Backpex.Fields.Select,
+            label: "Country",
+            prompt: "---",
```
Flo0807:
```suggestion
            prompt: "—",
```

lib/backpex/fields/belongs_to.ex
```diff
@@ -130,12 +130,13 @@ defmodule Backpex.Fields.BelongsTo do
       assigns
       |> assign(:options, options)
       |> assign(:owner_key, owner_key)
+      |> assign_new(:hide_label, fn -> false end)
```
Flo0807: I'd prefer not to set this assign in every `render_form/1` function. What are your thoughts on assigning it in `resource_field/1` and `inlined_resource_field/1`?

lib/backpex/fields/inline_crud.ex
```diff
@@ -7,7 +7,7 @@ defmodule Backpex.Fields.InlineCRUD do
     ],
     child_fields: [
       doc: """
-      A list of input fields to be used. Currently only support `Backpex.Fields.Text` fields.
+      A list of input fields to be used. Some fields, such as `Backpex.Fields.Upload`, are not supported.
```
Flo0807: I guess we should list all incompatible fields, not just examples.

lib/backpex/html/layout.ex
```diff
@@ -503,12 +503,20 @@ defmodule Backpex.HTML.Layout do
 
   attr :class, :any, default: nil, doc: "extra classes to be added"
 
-  slot :label, required: true do
+  slot :label, required: false do
```
Flo0807:
```suggestion
  slot :label do
```

lib/backpex/fields/belongs_to.ex
gmazzamuto: Yes it definitely makes sense, I'll review this PR asap

PR #1677, opened 2025-11-25
Support readonly for all fields

gmazzamuto:
Support `:readonly` option for all fields.

The `:readonly` field option, being shared among all fields, has been moved to the `@config_schema` of `Backpex.Field`. This should not be a breaking change.

Example of readonly fields:

Screenshot 2025-11-25 at 01-24-43 Edit User · Backpex


lib/backpex/html/core_components.ex
```diff
@@ -38,6 +38,7 @@ defmodule Backpex.HTML.CoreComponents do
       </.dropdown>
   """
   attr :id, :string, required: true, doc: "unique identifier for the dropdown"
+  attr :readonly, :boolean, default: false
```
Flo0807:
```suggestion
  attr :readonly, :boolean, default: false, doc: "whether the dropdown is readonly"
```

lib/backpex/html/form.ex
```diff
@@ -47,6 +47,13 @@ defmodule Backpex.HTML.Form do
   slot :inner_block
 
   def input(%{field: %Phoenix.HTML.FormField{} = field} = assigns) do
+    assigns =
+      if Map.has_key?(assigns.rest, :disabled) do
+        assigns
+      else
+        put_in(assigns, [:rest, :disabled], Map.get(assigns.rest, :readonly) || false)
+      end
```
Flo0807:
```suggestion
```

Why do we want to set `disabled` based on the `readonly` attribute? If you want to set the input to disabled, you should explicitly pass the `disabled` attribute, in my opinion.

lib/backpex/html/form.ex
```diff
@@ -205,6 +212,13 @@ defmodule Backpex.HTML.Form do
               multiple pattern placeholder readonly required rows size step)
 
   def currency_input(%{field: %Phoenix.HTML.FormField{} = field} = assigns) do
+    assigns =
+      if Map.has_key?(assigns.rest, :disabled) do
+        assigns
+      else
+        put_in(assigns, [:rest, :disabled], Map.get(assigns.rest, :readonly) || false)
+      end
```
Flo0807: See https://github.com/naymspace/backpex/pull/1677/files#r2591617008

lib/backpex/html/core_components.ex
```diff
@@ -64,21 +65,38 @@ defmodule Backpex.HTML.CoreComponents do
...
+      if assigns.readonly do
+        ["cursor-not-allowed bg-base-200"] ++
+          (trigger_class
+           |> Enum.join(" ")
+           |> String.split()
+           |> List.delete("bg-transparent")
+           |> List.delete("input"))
+      else
+        trigger_class
+      end
```
Flo0807: This looks complicated. Can't we just add additional classes in the markup in case the dropdown is readonly?

lib/backpex/html/form.ex
Flo0807: See https://github.com/naymspace/backpex/blob/0.16.3/lib/backpex/fields/text.ex#L58

I noticed that we do not set `dsiabled` in every field (e.g. `Backpex.Fields.Select`)

lib/backpex/fields/multi_select.ex
```diff
@@ -161,6 +161,7 @@ defmodule Backpex.Fields.MultiSelect do
         </:label>
         <Form.multi_select
           field={@form[@name]}
+          readonly={@readonly}
```
Flo0807, line 165: > undefined attribute "readonly" for component Backpex.HTML.Form.multi_select/1

lib/backpex/html/form.ex
```diff
@@ -286,6 +300,7 @@ defmodule Backpex.HTML.Form do
   @doc type: :component
 
   attr :prompt, :string, required: true, doc: "string that will be shown when no option is selected"
+  attr :readonly, :boolean, default: false
```
Flo0807:
```suggestion
  attr :readonly, :boolean, default: false, doc: "whether the dropdown is readonly"
```

lib/backpex/html/resource.ex
```diff
@@ -175,7 +175,13 @@ defmodule Backpex.HTML.Resource do
...
       |> assign(:field_options, field_options)
       |> assign(:type, :form)
-      |> assign(:readonly, Backpex.Field.readonly?(field_options, assigns))
+
+    assigns =
+      if assigns[:readonly] == true do
+        assigns
+      else
+        assign(assigns, :readonly, Backpex.Field.readonly?(field_options, assigns))
+      end
```
Flo0807, line 195: For which cases do we need this?

And can we simplify the if-condition?


```suggestion
    assigns =
      if assigns[:readonly] do
        assigns
      else
        assign(assigns, :readonly, Backpex.Field.readonly?(field_options, assigns))
      end
```

lib/backpex/html/form.ex
gmazzamuto: yeah you're right, I'll review your comments to this PR asap (some time has passed and I can't remember right now). Anyways I was confused by the fact that in Backpex almost all fields have
```html
<BackpexForm.input
  ...
  readonly={@readonly}
  disabled={@readonly}
>
```
Do you think it makes sense to have two separate options, `:readonly` and `:disabled`? After all, they serve different purposes. If that's the case I can work on that and add the `:disabled` option to `Backpex.Field`.

lib/backpex/html/core_components.ex
gmazzamuto: Here is why. This is how the dropdown is rendered with the code above:

Screenshot_20251223_173629

---

Whereas this code:
```elixir
    trigger_class =
      if assigns.readonly do
        ["cursor-not-allowed bg-base-200"] ++ trigger_class
      else
        trigger_class
      end
```

results in this, and the field can be even highlighted on mouse click (although the dropdown does not open):

Screenshot_20251223_173556

lib/backpex/html/resource.ex
gmazzamuto: Sure, the if condition can be simplified.

The reason for this code is because the `:readonly` assign is not always defined when calling the callback `Backpex.Field.render_form/1` and also to allow individual fields in `Backpex.Fields.InlineCRUD` to be set readonly when the parent `InlineCRUD` field is not readonly. I will add a comment in the code to clarify this.

gmazzamuto:
Hi there, I have merged the latest develop branch and added an extra commit to fix a bug.

What is left to do is to homogenize the look and feel of disabled fields. At the moment, disabled text fields look differently than other fields such as Select. I see that you are favoring the `readonly` attribute in place of `disabled` for accessibility issues and also because they have different meanings. This makes sense, but DaisyUI prefers `disabled` instead. See [here](https://daisyui.com/components/input/#disabled). Maybe we'll have to find another way to visually render a field as readonly/disabled.

lib/backpex/html/form.ex
Flo0807: Addressed: the shared form inputs no longer infer `disabled` from `readonly`. Text-like fields use native `readonly`, while Select, BelongsTo, and Boolean explicitly pass `disabled` because those controls do not support native readonly behavior. Commit 8ced2b48 gives these fields consistent theme-aware styling while preserving focus and text selection for readonly text inputs.

Regarding a separate field-level `disabled` option: let's keep that out of this PR. The existing `readonly` option expresses that the value cannot be edited, and each field chooses the appropriate HTML behavior. A separate public option would need its own defined submission and focus semantics.

lib/backpex/html/form.ex
Flo0807: Addressed here as well: the shared input component preserves the explicitly supplied `readonly` and `disabled` attributes independently. Readonly text fields remain focusable and selectable; control-style fields explicitly pass `disabled`. Their appearance is now consistent through the shared styles in 8ced2b48.

lib/backpex/html/core_components.ex
Flo0807: The visual treatment now uses shared conditional utility classes (8ced2b48). We retained separate readonly dropdown markup because the behavior also changes: readonly mode omits the menu and dropdown hook, removes the button role and tab stop, and renders a labelled group containing the selected values. Changing CSS classes alone would leave those interactive semantics in place.

This also preserves the field's accessible name. Verified the appearance in light/dark themes at desktop, tablet, and mobile widths, along with keyboard navigation.

lib/backpex/fields/multi_select.ex
Flo0807: Fixed: `Backpex.HTML.Form.multi_select/1` declares `attr :readonly, :boolean, default: false` and forwards it to the dropdown and selected badges. The undefined-attribute warning is gone. Library and demo compilation/lint passed, along with 614 library tests/doctests and 13 product/user form integration tests.

lib/backpex/html/resource.ex
Flo0807: Applied the suggested simplified condition and added the InlineCRUD explanation. An inherited `readonly: true` keeps all child fields readonly; otherwise each child's own readonly configuration is evaluated. This supports individually readonly children inside an editable InlineCRUD field.

We also fixed the nested Boolean submission issue in c3945850: disabling the toggle now disables its hidden false fallback too, so saving an unrelated field preserves an existing readonly true value. The regression test reproduces the old data loss and passes with the fix.

PR #1778, opened 2026-01-28
Add readonly support to Backpex.Fields.Currency

kepi:
I believe there is no actual reason not to use readonly for Currency fields too?


Flo0807:
Closed in favor of https://github.com/naymspace/backpex/pull/1677

PR #1779, opened 2026-01-28
Supress IgniterJs warnings when not installed

kepi:
Honestly, I'm completely unsure if this is correct way or if this should be merged. My only motivation is that I can't work on backpex without it as every file I edit in my project gets "==> backpex" at the top from `mix format` 🥹

I kind of thing that this might be bug in Mix, but as complete Elixir noob, I might be completely wrong.

Anyway, as IgniterJs is optional dependency I believe we shouldn't send warnings warnigns to output when it is missing.

Feel free to close this without merging, I'l simply leave it locally or install IgniterJs in my project. I just thought it might be helpful to somebody else.

### DIFF
--- a/lib/mix/tasks/backpex.install.ex
+++ b/lib/mix/tasks/backpex.install.ex
@@ -61,6 +61,11 @@ if Code.ensure_loaded?(Igniter) do
     alias Igniter.Project.Module
     alias Igniter.Util.IO, as: IgniterIO
     alias Igniter.Util.Warning
+
+    # IgniterJs is an optional dependency - suppress warnings when not installed
+    @compile {:no_warn_undefined, {IgniterJs.Helpers, :read_and_validate_file, 1}}
+    @compile {:no_warn_undefined, {IgniterJs.Parsers.Javascript.Parser, :insert_imports, 3}}
+    @compile {:no_warn_undefined, {IgniterJs.Parsers.Javascript.Parser, :extend_hook_object, 3}}
     alias IgniterJs.Parsers.Javascript.Parser
 
     @default_app_css_path Path.join(["assets", "css", "app.css"])


lib/mix/tasks/backpex.install.ex
```diff
@@ -61,6 +61,11 @@ if Code.ensure_loaded?(Igniter) do
     alias Igniter.Project.Module
     alias Igniter.Util.IO, as: IgniterIO
     alias Igniter.Util.Warning
+
+    # IgniterJs is an optional dependency - suppress warnings when not installed
+    @compile {:no_warn_undefined, {IgniterJs.Helpers, :read_and_validate_file, 1}}
+    @compile {:no_warn_undefined, {IgniterJs.Parsers.Javascript.Parser, :insert_imports, 3}}
+    @compile {:no_warn_undefined, {IgniterJs.Parsers.Javascript.Parser, :extend_hook_object, 3}}
```
Flo0807, line 68: Rather than suppressing the warning, we should fix the root cause.

Although we already check whether the Igniter is installed using `Code.ensure_loaded?(Igniter)` (see https://github.com/naymspace/backpex/blob/0.17.1/lib/mix/tasks/backpex.install.ex#L48), the intall task depends on both the `igniter` and `igniter_js`. If you have installed `igniter`, but not `igniter_js`, the installation command is compiled and a warning is logged because `igniter_js` is missing.

I suggest that we only load the install MIx task if both `igniter` and `igniter_js` are installed.

Flo0807 (changes requested):

lib/mix/tasks/backpex.install.ex
kepi: Definitely agree. If I don't understand it wrong, II suppose you can close this PR as it is no longer needed. Thanks.

Edit: sorry, just realized I can close it myself...

Flo0807: Yes, I fixed this in #1821.

Anyway, thanks for the contribution. Otherwise, we would not have noticed this issue!

PR #1780, opened 2026-01-28
Fix dropping readonly fields from change

kepi:
When experimenting with AshBackpex, I found out that :readonly fields aren't correctly dropped from a Change.

As I have :update action which accepts only some of fields, I need this working. Only alternative would be to omit read only fields from edit form which doesn't suit my use case.

I also added simple tests for this, but I don't really know how to test this correctly with existing tests structure. For now, I used dirty workaround and copied private `drop_readonly_changes` function directly to tests, just so I can test if it works.

But I think better (not ideal) way would be making the `drop_readonly_changes` non-private and add `@doc false`. Are you OK with that? Ideal would be probably to setup some "infrastructure" for LiveView and check whole process, but I'm not even sure where to begin with that.

### DIFF
--- a/lib/backpex/field.ex
+++ b/lib/backpex/field.ex
@@ -357,6 +357,21 @@ defmodule Backpex.Field do
   def readonly?(%{readonly: readonly}, assigns) when is_function(readonly, 1), do: readonly.(assigns)
   def readonly?(_field_options, _assigns), do: false
 
+  @doc """
+  Drops readonly field changes from the given change map.
+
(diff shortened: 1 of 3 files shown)


Flo0807:
> But I think better (not ideal) way would be making the drop_readonly_changes non-private and add @doc false. Are you OK with that? Ideal would be probably to setup some "infrastructure" for LiveView and check whole process, but I'm not even sure where to begin with that.

I moved the function to `Backpex.Field` and made it public.

PR #1783, opened 2026-02-01
Show validation errors for `Backpex.Fields.InlineCRUD`

sebasgo:
... not only for the child fields, but also for the main field itself. This is useful to validate constraints over all associated items at once. 

I have tried to make the error handling similar to the other fields, but found the complex ones which do not delegate this to a form widget a bit inconsistent with regard to error handling. Feel free to correct my approach here.

### DIFF
--- a/lib/backpex/fields/inline_crud.ex
+++ b/lib/backpex/fields/inline_crud.ex
@@ -104,8 +104,7 @@ defmodule Backpex.Fields.InlineCRUD do
   end
 
   defp apply_action(socket, :form) do
-    socket
-    |> assign_form_errors()
+    assign_form_errors(socket)
   end
 
   defp apply_action(socket, _type), do: socket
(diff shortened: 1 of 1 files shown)


lib/backpex/fields/inline_crud.ex
```diff
@@ -98,9 +99,17 @@ defmodule Backpex.Fields.InlineCRUD do
     socket
     |> assign(assigns)
     |> assign(child_fields: child_fields)
+    |> apply_action(assigns.type)
     |> ok()
   end
 
+  defp apply_action(socket, :form) do
+    socket
+    |> assign_form_errors()
```
Flo0807:
```suggestion
    assign_form_errors(socket)
```

lib/backpex/fields/inline_crud.ex
```diff
@@ -98,9 +99,17 @@ defmodule Backpex.Fields.InlineCRUD do
...
+  end
+
+  defp apply_action(socket, _), do: socket
```
Flo0807:
```suggestion
  defp apply_action(socket, _type), do: socket
```

Flo0807:
Thanks for the contribution! 🙌

PR #1808, opened 2026-02-11
Allow defining layout in callback function

mitchellhenke:
Hello!

I was debugging some [compile cycles](https://hexdocs.pm/mix/main/Mix.Tasks.Xref.html) in a Phoenix application, and noticed that `backpex` was part of the issue.

The demo application has a good example of what I'm running into. If you run `mix xref graph --format cycles --label compile-connected`, it reports that there is a compile cycle:

```
demo/ $ mix xref graph --format cycles --label compile-connected
1 cycles found. Showing them in decreasing size:

Cycle of length 16 (9 compile, 1 export):

    lib/demo_web/components/core_components.ex
    lib/demo_web/components/layouts.ex (export)
    lib/demo_web/controllers/redirect_controller.ex
    lib/demo_web/endpoint.ex
    lib/demo_web/item_actions/user_soft_delete.ex
    lib/demo_web/live/address_live.ex (compile)
    lib/demo_web/live/category_live.ex (compile)
    lib/demo_web/live/film_review_live.ex (compile)
    lib/demo_web/live/invoice_live.ex (compile)
    lib/demo_web/live/post_live.ex (compile)
    lib/demo_web/live/product_live.ex (compile)
    lib/demo_web/live/short_link_live.ex (compile)
    lib/demo_web/live/tag_live.ex (compile)
    lib/demo_web/live/user_live.ex (compile)
    lib/demo_web/resource_actions/upload.ex
    lib/demo_web/router.ex
```

This results in a change in any of the files recompiling all of the files, which usually means slower compile times during development in apps that use Backpex.

I'm not deeply familiar with Elixir compiler functionality, but my understanding is the reason for it is the `DemoWeb.Layouts` becomes a compile dependency of the LiveViews when specified in `use Backpex.LiveResource`. `DemoWeb.Layouts` includes `use DemoWeb, :html` which ends up pulling in the router, etc.

To continue the above example using the demo app, if you apply [this patch](https://github.com/user-attachments/files/25243387/fix_compile_cycle.patch) to replace the layout option with the callback, the demo app no longer has a compile cycle:

```
demo/ $ mix xref graph --format cycles --label compile-connected
No cycles found
```

I'm proposing allowing specifying the layout as a callback function to avoid the specific compile dependency between layouts and the LiveView. To avoid a breaking change, I've not removed the current option. I also haven't tried updating any guides, documentation, or similar, but am happy to if you're open to this change.

### DIFF
--- a/demo/lib/demo_web/live/address_live.ex
+++ b/demo/lib/demo_web/live/address_live.ex
@@ -6,8 +6,10 @@ defmodule DemoWeb.AddressLive do
       update_changeset: &Demo.Address.update_changeset/3,
       create_changeset: &Demo.Address.create_changeset/3
     ],
-    layout: {DemoWeb.Layouts, :admin},
     fluid?: true
+
+  @impl Backpex.LiveResource
+  def layout(_assigns), do: {DemoWeb.Layouts, :admin}
 
   @impl Backpex.LiveResource
   def singular_name, do: "Address"
--- a/demo/lib/demo_web/live/category_live.ex
+++ b/demo/lib/demo_web/live/category_live.ex
@@ -6,8 +6,10 @@ defmodule DemoWeb.CategoryLive do
       update_changeset: &Demo.Category.update_changeset/3,
       create_changeset: &Demo.Category.create_changeset/3
     ],
-    layout: {DemoWeb.Layouts, :admin},
     init_order: %{by: :name, direction: :asc}
+
+  @impl Backpex.LiveResource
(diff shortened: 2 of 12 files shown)


Flo0807:
Hey, thanks for your contribution!

mitchellhenke:
> Hey, thanks for your contribution!

Of course, thanks for the very helpful library! Happy to help with getting this merged, just let me know 🙂

pehbehbeh (approved):
LGTM

Maybe even soft deprecate the old way via NimbleOptions (see `:deprecated` option)?

PR #1855, opened 2026-03-06
Add optional `link/2` callback to ItemActions

### LINKED ISSUE #1854: Item actions as links
### Discussed in https://github.com/naymspace/backpex/discussions/1815

<sup>Originally posted by **dorianamouroux** February 14, 2026</sup>
Hey!

First off, thanks for Backpex, we're using it for our admin and it's been really smooth.

One thing we've run into: item actions are always rendered as `<button>` with `phx-click="item-action"`, so every click goes through the server. For simple navigation actions (e.g. "Show", "Edit", or custom links), that's a bit heavy.

We'd love to have an option for link-style item actions that render as real `<a href="...">` instead of buttons.

## Why we'd find it useful:
 - **Faster navigation**: no LiveView round-trip for "just go to this URL".
 - **Better browser UX**: Ctrl+click / Cmd+click to open in a new tab, right-click "Open in new tab", etc.

## Rough idea:

Something like a type: :link (or similar) on the action config: where the action module provides a way to compute the URL (e.g. a path(socket, item) callback or the like) instead of (or in addition to) handle/3:

```elixir
defmodule ExampleWeb.Live.Admin.BackpexItemActions.ViewAction do
  use BackpexWeb, :item_action, type: :link

  @impl Backpex.ItemAction
  def icon(assigns, _item) do
    ~H"""
    <Backpex.HTML.CoreComponents.icon
      name="hero-check-circle"
      class="h-5 w-5 cursor-pointer transition duration-75 hover:text-success hover:scale-110"
    />
    """
  end

  @impl Backpex.ItemAction
  def label(_assigns, _item), do: "Approve"

  @impl Backpex.ItemAction
  def href(socket, item, _data) do
     path = Router.get_path(socket, item.live_resource, Map.get(assigns, :params, %{}), :show, item)
  end
end
```

For those actions, the template would render an <a href="..."> with the same icon/label/tooltip, and skip the phx-click="item-action" flow.

No strong opinion on the exact API — just wanted to float the idea and see what you think. Happy to help with a PR if you're open to it.

Cheers!
</div>

Flo0807:
closes #1854 

Item actions can now optionally implement a link/2 callback that returns a URL path. When present, the action renders as a <.link navigate={path}> instead of a <button>, enabling Ctrl+click to open in new tab and right-click context menus. Built-in Show and Edit actions implement link/2 by default. Fully backwards-compatible — existing actions without link/2 continue to render as buttons.

### DIFF
--- a/lib/backpex/html/resource/resource_index_table.html.heex
+++ b/lib/backpex/html/resource/resource_index_table.html.heex
@@ -66,9 +66,20 @@
         "text-base-content/75 z-10 w-0 px-6"
       ]}>
         <div class={["flex items-center justify-end space-x-2"]}>
+          <.link
+            :for={{key, action} <- filter_item_actions(@item_actions, :row)}
+            :if={Backpex.ItemAction.has_link?(action) and @live_resource.can?(assigns, key, item)}
+            id={"item-action-#{key}-#{LiveResource.primary_value(item, @live_resource)}"}
+            navigate={action.module.link(assigns, item)}
+            aria-label={action.module.label(assigns, item)}
+            phx-hook="BackpexTooltip"
+            data-tooltip={action.module.label(assigns, item)}
+          >
+            {action.module.icon(assigns, item)}
+          </.link>
           <button
             :for={{key, action} <- filter_item_actions(@item_actions, :row)}
-            :if={@live_resource.can?(assigns, key, item)}
+            :if={not Backpex.ItemAction.has_link?(action) and @live_resource.can?(assigns, key, item)}
             id={"item-action-#{key}-#{LiveResource.primary_value(item, @live_resource)}"}
             type="button"
             phx-click="item-action"
--- a/lib/backpex/html/resource/resource_index_table.html.heex
+++ b/lib/backpex/html/resource/resource_index_table.html.heex
@@ -66,31 +66,33 @@
         "text-base-content/75 z-10 w-0 px-6"
       ]}>
         <div class={["flex items-center justify-end space-x-2"]}>
-          <.link
-            :for={{key, action} <- filter_item_actions(@item_actions, :row)}
-            :if={Backpex.ItemAction.has_link?(action) and @live_resource.can?(assigns, key, item)}
-            id={"item-action-#{key}-#{LiveResource.primary_value(item, @live_resource)}"}
-            navigate={action.module.link(assigns, item)}
-            aria-label={action.module.label(assigns, item)}
-            phx-hook="BackpexTooltip"
-            data-tooltip={action.module.label(assigns, item)}
-          >
-            {action.module.icon(assigns, item)}
-          </.link>
-          <button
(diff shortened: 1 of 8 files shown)


lib/backpex/html/resource/resource_index_table.html.heex
```diff
@@ -66,9 +66,20 @@
         "text-base-content/75 z-10 w-0 px-6"
       ]}>
         <div class={["flex items-center justify-end space-x-2"]}>
+          <.link
```
dorianamouroux: As we are rendering first the links, then the buttons; we are not fully respecting the order in which actions are defined in the resource.

If that's a concern, perhaps we could have an approach like so instead:

```elixir
<%= for {key, action} <- filter_item_actions(@item_actions, :row),
        @live_resource.can?(assigns, key, item) do %>

  <%= if Backpex.ItemAction.has_link?(action) do %>
    <.link
      id={"item-action-#{key}-#{LiveResource.primary_value(item, @live_resource)}"}
      navigate={action.module.link(assigns, item)}
      aria-label={action.module.label(assigns, item)}
      phx-hook="BackpexTooltip"
      data-tooltip={action.module.label(assigns, item)}
    >
      {action.module.icon(assigns, item)}
    </.link>
  <% else %>
    <button
      id={"item-action-#{key}-#{LiveResource.primary_value(item, @live_resource)}"}
      type="button"
      phx-click="item-action"
      phx-value-action-key={key}
      phx-value-item-id={LiveResource.primary_value(item, @live_resource)}
      aria-label={action.module.label(assigns, item)}
      phx-hook="BackpexTooltip"
      data-tooltip={action.module.label(assigns, item)}
    >
      {action.module.icon(assigns, item)}
    </button>
  <% end %>

<% end %>
```

WDYT?

Flo0807: You're right. We should preserve the order!

lib/backpex/item_actions/item_action.ex
```diff
@@ -163,6 +175,39 @@ defmodule Backpex.ItemAction do
...
+          file: env.file,
+          line: env.line,
+          description: """
+          ItemAction #{inspect(env.module)} implements both handle/3 and link/2.
+
+          An item action must implement exactly one of these callbacks.
+          Use handle/3 for actions that perform server-side operations,
+          or link/2 for actions that navigate to a URL.
+          """
+
```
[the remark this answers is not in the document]
Flo0807: Re: allowing both `handle/3` and `link/2` — the mutual exclusivity is intentional. If `link/2` is defined, `handle/3` would be dead code since link-based actions navigate directly via the rendered anchor tag without a server round-trip. Allowing both would silently leave dead code in the module. The `:def` check (cde16d3f) ensures private helpers named `link` are not caught by this validation.

pehbehbeh (approved):
LGTM

PR #1856, opened 2026-03-06
Add optional view link to InlineCRUD field rows

### LINKED ISSUE #1857: Auto-generate link field in InlineCRUD when LiveResource is provided
### Discussed in https://github.com/naymspace/backpex/discussions/1818

<sup>Originally posted by **dorianamouroux** February 18, 2026</sup>
Hi! 👋

I've been using Backpex and loving it so far. I have a small feature suggestion that would make working with `InlineCRUD` a bit smoother.

## Problem

Currently, linking to a resource from `Backpex.Fields.InlineCRUD` requires manually creating a custom link field with routing logic:

```elixir
books: %{
  module: Backpex.Fields.InlineCRUD,
  type: :assoc,
  child_fields: [
    id: %{
      module: LinkTo,  # Custom field
      label: "View",
      only: [:show],
      href: fn assigns ->
        Router.get_path(assigns.socket, Book, %{}, :show, assigns.item)
      end
    }
  ]
}
```

## Proposed Solution

Add a `live_resource` option to `InlineCRUD` (similar to `BelongsTo`). When provided, automatically generate a link field:

```elixir
product_options: %{
  module: Backpex.Fields.InlineCRUD,
  type: :assoc,
  live_resource: LibraryWeb.Live.Admin.Book,
  child_fields: [
    # ... other fields ...
  ]
}
```

So in a resource A (let's say Author) we can have an InlineCRUD to easily add another resource (e.g. books). But we also want a resource definition for the Book to add extra details we don't want in the InlineCrud field.

---

If this sounds like a good addition, I'd be happy to take on implementing this feature! Let me know what you think. 🙂
</div>

Flo0807:
closes #1857 

When a live_resource option is provided on an InlineCRUD field, each row in the read-only table view gets an eye icon link that navigates to the associated resource's show page. This mirrors the existing pattern used by BelongsTo and HasMany fields.

The link includes a tooltip and respects authorization via can?/3. When live_resource is not set, no action column is rendered, keeping the feature fully backwards-compatible.

image

### DIFF
--- a/demo/lib/demo_web/live/product_live.ex
+++ b/demo/lib/demo_web/live/product_live.ex
@@ -124,6 +124,7 @@ defmodule DemoWeb.ProductLive do
         label: "Short Links",
         type: :assoc,
         except: [:index],
+        live_resource: DemoWeb.ShortLinkLive,
         child_fields: [
           short_key: %{
             module: Backpex.Fields.Text,
--- a/lib/backpex/fields/inline_crud.ex
+++ b/lib/backpex/fields/inline_crud.ex
@@ -19,6 +19,11 @@ defmodule Backpex.Fields.InlineCRUD do
       """,
       type: :keyword_list,
       required: true
+    ],
+    live_resource: [
+      doc:
+        "The live resource of the association. When provided, a link to each item's show page is rendered in the read-only view.",
+      type: :atom
     ]
   ]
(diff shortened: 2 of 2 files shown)


pehbehbeh (approved):
LGTM

PR #1862, opened 2026-03-10
Refactor release reset to drop schema instead of migrating down

Flo0807:
`migrate(:down)` fails when migration files have been deleted. Replace it with `DROP SCHEMA public CASCADE` to reliably wipe all tables. Also fix incorrect `:myxql` reference to :postgrex in `seed/0`.


pehbehbeh (approved):
LGTM

PR #1898, opened 2026-04-04
Fix: make on_item_created/updated/deleted overridable without compiler warnings

vanderhoop:
## Summary

Moves `on_item_created/2`, `on_item_updated/2`, and `on_item_deleted/2` default implementations from `__before_compile__` into `__using__` and adds them to the `defoverridable` list.

## Problem

When overriding these callbacks in a LiveResource module, the Elixir compiler emits a "redundant clause" warning:

```
warning: the following clause is redundant:

    def on_item_updated(socket, _item)

it has type:

    dynamic(), dynamic()

previous clauses have already matched on the following types:

    term(), term()
```

This happens because the defaults are defined in `__before_compile__`, which runs **after** the module body. The user's override becomes an earlier clause matching `(term(), term())`, making Backpex's default clause unreachable — correct behavior, but with a noisy warning.

## Fix

Move the three `on_item_*` defaults into the `__using__` block (alongside the other defaults like `can?/3`, `fields/0`, etc.) and add them to the existing `defoverridable` call. This lets users override them cleanly without compiler warnings.

## Test plan

- Verified in a consuming Phoenix app that overriding `on_item_updated/2` no longer produces the warning
- Full test suite passes with no regressions

### DIFF
--- a/lib/backpex/live_resource.ex
+++ b/lib/backpex/live_resource.ex
@@ -320,6 +320,15 @@ defmodule Backpex.LiveResource do
       @impl Backpex.LiveResource
       def item_actions(default_actions), do: default_actions
 
+      @impl Backpex.LiveResource
+      def on_item_created(socket, _item), do: socket
+
+      @impl Backpex.LiveResource
+      def on_item_updated(socket, _item), do: socket
+
+      @impl Backpex.LiveResource
+      def on_item_deleted(socket, _item), do: socket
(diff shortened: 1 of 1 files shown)


vanderhoop:
Closing, as a pattern match on the struct passed as the second argument in these callbacks results in the specificity needed 🤦🏻‍♂️

PR #1921, opened 2026-04-15
Enable :module_directives rule in Quokka

pehbehbeh:
###  Summary

  Enables Quokka's `:module_directives` rule project-wide (previously excluded) so module directives get organized consistently. Files that intentionally define `@config_schema` before `@moduledoc` (to interpolate
  schema docs into the moduledoc) are opted out with `# quokka:skip-module-directive-reordering`.

###  Changes

  - Remove `exclude: [:module_directives]` from the root and demo `.formatter.exs`.
  - Add skip comments to 19 field/adapter modules that must keep `@config_schema` before `@moduledoc`.
  - Manually hoist `Phoenix.LiveView.Rendered` / `Socket` aliases in `Backpex.Field` so `@callback` specs resolve under the preserved source order.
  - Apply the broader `# quokka:skip-module-directives` to `post_live.ex` and `product_live.ex` in the demo, where Quokka's alias lifting of `Backpex.Fields.Number` shadowed the `Number` Hex package used by
  `Number.Delimit`.
  - Absorb the resulting reformatting across demo modules and tests.

### DIFF
--- a/demo/lib/demo/invoice.ex
+++ b/demo/lib/demo/invoice.ex
@@ -3,12 +3,14 @@ defmodule Demo.Invoice do
 
   use Ecto.Schema
 
+  alias Money.Ecto.Amount.Type
+
   @primary_key {:id, :binary_id, autogenerate: true}
 
   schema "invoices" do
     field :company, :string
 
-    field :amount, Money.Ecto.Amount.Type
+    field :amount, Type
 
     timestamps()
   end
--- a/demo/lib/demo/invoice.ex
+++ b/demo/lib/demo/invoice.ex
@@ -3,14 +3,12 @@ defmodule Demo.Invoice do
 
   use Ecto.Schema
 
-  alias Money.Ecto.Amount.Type
-
   @primary_key {:id, :binary_id, autogenerate: true}
 
   schema "invoices" do
     field :company, :string
 
-    field :amount, Type
+    field :amount, Money.Ecto.Amount.Type
 
     timestamps()
   end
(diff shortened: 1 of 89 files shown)


pehbehbeh:
@Flo0807 FYI: We can further improve this once https://github.com/emkguts/quokka/pull/150 gets merged.

demo/lib/demo_web/live/invoice_live.ex
Flo0807, line 1: Some LiveResources use `# quokka:skip-module-directives` some not.

demo/lib/demo/invoice.ex
```diff
@@ -3,12 +3,14 @@ defmodule Demo.Invoice do
...
 
+  alias Money.Ecto.Amount.Type
+
   @primary_key {:id, :binary_id, autogenerate: true}
 
   schema "invoices" do
     field :company, :string
 
-    field :amount, Money.Ecto.Amount.Type
+    field :amount, Type
```
Flo0807: Personally, I'd prefer
```elixir
field :amount, Money.Ecto.Amount.Type
```
but let's not make exceptions here and there

pehbehbeh: Quokka should respect `Credo.Check.Design.AliasUsage` which has a few settings for finetuning. WDYT?

https://hexdocs.pm/credo/Credo.Check.Design.AliasUsage.html

I can try that before merging...

Flo0807: yes, let's try

Flo0807 (approved):

PR #1936, opened 2026-04-23
Upgrade Yarn from 1 to 4

Flo0807:
Manage yarn via Corepack using the packageManager field in both package.json files. Use nodeLinker: node-modules for Phoenix/esbuild compatibility. Quote glob patterns in npm-run-all2 scripts since Yarn 4's shell rejects unmatched bare globs.

- Add .yarnrc.yml to root and demo
- Replace --pure-lockfile with --immutable in Dockerfile, compose.yml, and CI workflows
- Enable Corepack in Dockerfile and CI workflows
- Regenerate yarn.lock in Yarn 4 format
- Ignore Yarn 4 artifacts (.yarn/*, .pnp.*)


Flo0807:
Switching from Yarn 4 migration to Bun migration instead.

PR #2170, opened 2026-08-07
Allow app web helpers in Backpex extensions

Flo0807:
## Summary

- import Phoenix component helpers for item actions, filters, and metrics without registering a second declarative compile hook
- allow custom fields to use their application's LiveComponent entrypoint via `live_component: {MyAppWeb, :live_component}`
- keep full `Phoenix.Component` setup in Backpex's built-in filters and value metric because they declare `attr`
- document application Gettext/component/verified-route usage and the upgrade impact
- add compile fixtures covering both `use` orders and host helpers under compiler verification

## Why

PR #2055 fixed duplicate `Phoenix.Component.Declarative` `@before_compile` hooks for LiveResources. The same duplicate hook was still introduced by `BackpexWeb, :item_action`, `:filter`, and `:metric` when consumers also used `MyAppWeb, :html`.

Fields need a different path because they are real LiveComponents. The new option lets the host application's standard `:live_component` entrypoint perform the LiveComponent setup exactly once while Backpex adds only its field-specific aliases and helpers.

## User impact

Item actions, custom filters, typed filters, and metrics may now use `MyAppWeb, :html` in either order to access application helpers. Extensions that declare their own `attr` or `slot` must explicitly use their application HTML entrypoint or `Phoenix.Component`.

Custom fields can use:

```elixir
use Backpex.Field,
  config_schema: @config_schema,
  live_component: {MyAppWeb, :live_component}
```

Existing fields without this option keep the current Backpex LiveComponent setup.

## Checks

- `MIX_ENV=test mix lint` — 275 tests passed (100 doctests, 175 tests), Credo clean
- `mix test test/backpex/web_helpers_test.exs` — passed with no compile or runtime diagnostics

### DIFF
--- a/guides/fields/custom-fields.md
+++ b/guides/fields/custom-fields.md
@@ -53,6 +53,18 @@ The `render_form/1` function returns markup that is used to render a form on `ed
 
 See `Backpex.Field` for more information on the available callback functions. For example, you can implement `render_index_form/1` to make the field editable in the index view.
 
+## Using your application's web helpers
+
+Custom fields are LiveComponents. If a field needs helpers from your application, such as Gettext, core components, or verified routes, configure your application's LiveComponent entrypoint:
+
+```elixir
+use Backpex.Field,
+  config_schema: @config_schema,
+  live_component: {MyAppWeb, :live_component}
+```
+
+The standard Phoenix-generated `MyAppWeb, :live_component` entrypoint includes the application's HTML helpers while setting up `Phoenix.LiveComponent` exactly once. Do not additionally `use MyAppWeb, :html` in the same field module.
+
 ## Add field option validation
 
 With Backpex v0.9 we are validating field options. This ensures that only field options that are actually used by the field can be defined in the field options map. So if your custom field requires certain field options, make sure you define them.
--- a/guides/upgrading/v0.19.md
+++ b/guides/upgrading/v0.19.md
@@ -68,8 +68,8 @@ end
 
 The `BackpexWeb` entrypoints for item actions, filters, and metrics now import
 `Phoenix.Component` instead of using it. This avoids registering the
-`Phoenix.Component.Declarative` `@before_compile` hook twice when an extension also
-uses its application's HTML entrypoint.
+declarative `@before_compile` hook twice when an extension also uses its application's
+HTML entrypoint.
 
 You can now use application-level Gettext, components, and verified routes directly in
 these extensions, in either order:
--- a/guides/upgrading/v0.19.md
+++ b/guides/upgrading/v0.19.md
@@ -63,3 +63,47 @@ defmodule MyAppWeb.PostLive do
   end
 end
 ```
+
+## Backpex extension modules now import `Phoenix.Component`
+
+The `BackpexWeb` entrypoints for item actions, filters, and metrics now import
+`Phoenix.Component` instead of using it. This avoids registering the
+`Phoenix.Component.Declarative` `@before_compile` hook twice when an extension also
+uses its application's HTML entrypoint.
+
+You can now use application-level Gettext, components, and verified routes directly in
+these extensions, in either order:
+
+```elixir
+defmodule MyAppWeb.ItemActions.Archive do
+  use BackpexWeb, :item_action
+  use MyAppWeb, :html
+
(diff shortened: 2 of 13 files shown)


guides/upgrading/v0.19.md
```diff
@@ -63,3 +63,47 @@ defmodule MyAppWeb.PostLive do
   end
 end
 ```
+
+## Backpex extension modules now import `Phoenix.Component`
```
pehbehbeh, line 67: Should move to 0.20

guides/fields/custom-fields.md
```diff
@@ -53,6 +53,18 @@ The `render_form/1` function returns markup that is used to render a form on `ed
...
 See `Backpex.Field` for more information on the available callback functions. For example, you can implement `render_index_form/1` to make the field editable in the index view.
 
+## Using your application's web helpers
+
+Custom fields are LiveComponents. If a field needs helpers from your application, such as Gettext, core components, or verified routes, configure your application's LiveComponent entrypoint:
+
+```elixir
+use Backpex.Field,
+  config_schema: @config_schema,
+  live_component: {MyAppWeb, :live_component}
```
pehbehbeh, line 63: Does this fail on compile or on render? It should fail on compile.

(To be honest, I’m not a big fan of the syntax. Feels very much like a workaround. But I can’t think of anything better at the moment.)

lib/backpex/field.ex
```diff
@@ -226,15 +226,35 @@ defmodule Backpex.Field do
...
+  The configured entrypoint must set up a `Phoenix.LiveComponent`, as the standard
+  Phoenix-generated `MyAppWeb, :live_component` entrypoint does.
   """
   defmacro __using__(opts) do
-    quote bind_quoted: [opts: opts] do
+    opts = Macro.expand(opts, __CALLER__)
+
+    if !Keyword.keyword?(opts) do
+      raise ArgumentError, "expected Backpex.Field options to be a keyword list, got: #{Macro.to_string(opts)}"
+    end
```
pehbehbeh, line 245: 🤖 

**What changed.** Before the PR, `Backpex.Field.__using__` passed the options through `quote bind_quoted:`, so whatever expression the caller wrote was evaluated inside the caller's module body at compile time. Any expression that produced a keyword list worked. The PR now inspects the options as syntax tree at macro-expansion time: it calls `Macro.expand` on them, requires the result to literally be a keyword list, pops `:live_component` out of it, and raises otherwise. `Macro.expand` only expands macros at the top level. It never evaluates attributes, variables, or function calls, so anything that isn't a written-out `[key: value, ...]` fails the check.

**Shapes that break now.** Each of these compiled fine before and now raises at compile time:

  - `use Backpex.Field, @opts` with the options stored in a module attribute. The AST expands to an attribute read, not a list. This is the case I verified.
  - `use Backpex.Field, Keyword.merge(@base, config_schema: ...)` or any function call producing the options.
  - `use Backpex.Field, opts` where `opts` is a variable bound earlier in the module body.
  - A wrapper macro in the app or a library that forwards options through `bind_quoted`. That pattern injects a variable at the inner `use` site, so a base module like this now
  fails:

  ```elixir
  defmacro __using__(opts) do
    quote bind_quoted: [opts: opts] do
      use Backpex.Field, opts   # opts is a variable here, not a literal list
    end
  end
  ```

  **Shapes that still work.** A literal keyword list whose values are expressions is fine, because only the outer list shape is inspected. `use Backpex.Field, config_schema: @config_schema` keeps working, as does a macro that expands directly into a keyword list. All built-in fields and the documented usage are in this group, so most users won't hit it.

  **Why it matters anyway.** The failure is a compile error with a misleading message. Someone with a shared base field module sees `expected Backpex.Field options to be a keyword list, got: Module.__get_attribute__(__MODULE__, :opts, 3, true)` after a routine upgrade, with no hint that the fix is to inline the options.

  **Fix.** Only pop `:live_component` when the options are a literal keyword list, and otherwise pass them through untouched with the default `use Phoenix.LiveComponent`. That restores the old behaviour for every non-literal case. The only limitation is that `:live_component` itself must be given literally, which is reasonable since it has to be resolved at compile time to emit a `use`.

PR #2220, opened 2026-08-26
Centralize authorization enforcement

pehbehbeh:
### Summary

 Enforces authorization centrally instead of relying on every LiveView caller to remember `can?/3` checks. All mutations in `Backpex.Resource` now gate before any user code runs, and item/resource actions are re-checked at execution time — closing an active bypass where the submitted action key was read from a forgeable DOM parameter. Unauthorized items in a selection now raise `Backpex.ForbiddenError` instead of being silently filtered.

### Changes

- New `Backpex.Authorization` module: `can?`/`can_all?` for UI preflight, `authorize!`/`authorize_all!` as the execution gate
- `Resource.insert/update/delete_all/update_all` authorize by default, with `authorization_action:` override and `authorize?: false` escape hatch (breaking signatures for `delete_all`/`update_all`)
- Client-supplied item ids and action keys are validated server-side; unknown values raise `NoResultsError`
- Bulk-action buttons are disabled for empty or partially unauthorized selections
- Upgrade guide `v0.21.md` plus updated authorization and action guides


lib/backpex/live_components/form_component.ex
```diff
@@ -355,45 +360,61 @@ defmodule Backpex.FormComponent do
...
+          action_to_confirm: action,
           return_to: return_to
         } = assigns
     } = socket
 
+    action_key = action.key
+
+    # Gate before any changeset work: permission may have been revoked, or the selection widened,
+    # while the modal was open.
+    Authorization.authorize_all!(live_resource, assigns, action_key, selected_items)
```
Flo0807: High: This execution gate authorizes `selected_items` cached when rows were selected, rather than freshly loaded records. `Index.update_item/2` refreshes `assigns.items` but leaves `selected_items` stale. If another actor changes a selected user from `role:user` to `role:admin` while the modal is open, this line still authorizes the old `role:user` snapshot; dispatch then reaches `update_all` with `authorize?: false`, and the Ecto adapter mutates the current row by primary key. This defeats the promised submit-time permission re-check. Please reload the target records immediately before the authoritative gate and dispatch the fresh records; ideally keep reload/lock, authorization, and mutation in the same transaction.

Flo0807:
> error: cannot find or invoke local is_non_struct_map/1 inside guards. Only macros can be invoked in a guards and they must be defined before their invocation. Called as: is_non_struct_map(assigns)

Should we drop support for Elixir 1.16?`is_non_struct_map/1` requires Elixir 1.17+ @pehbehbeh

Flo0807 (approved):

